Issue #34, opened 2018-02-11
[TI-26] Day report places hours and blocks on same line


taskwarrior:
_Jonathon Bly on 2016-07-26T03:17:29Z says:_ 

 The named blocks of time in the day report are written over the hours of the day. This problem is not present in the week report.

taskwarrior:
Migrated metadata:
```
Created: 2016-07-26T03:17:29Z
Modified: 2016-09-01T20:21:38Z
```

taskwarrior:
_Paul Beckingham on 2016-07-26T13:01:30Z says:_ 

 Is this just the effects of `rc.reports.day.axis`?  What does `timew get dom.rc.reports.day.axis` show?

taskwarrior:
_Paul Beckingham on 2016-08-16T20:30:02Z says:_ 

 No response.

taskwarrior:
_Jan Stolarek on 2016-09-01T20:16:08Z says:_ 

 I am also seeing this - please reopen. `timew get dom.rc.reports.day.axis` reports `internal`.

taskwarrior:
_Paul Beckingham on 2016-09-01T20:21:38Z says:_ 

 Right, you're seeing a feature.  Compare these:

```
$ timew week rc.reports.week.axis=internal
$ timew week rc.reports.week.axis=external
```

That feature only looks right for the `day` report, where it is the default.

PR #117, opened 2018-03-16
Docs: try to clarify reports.<type>.axis description

rudis:
Hello,

Like #34 I was confused about the differing behavior in `timew day`. This pull request trys to clarify the axis docs.

Regards
Simon


rudis:
The reason why I chose `<no value>` was that the code internally doesn't use `external` at all. It only knows `internal`. But I can see why this is confusing. Maybe like this?

rudis:
I've pushed another version which clarifies that in the future additional values might be defined. What do you think?

lauft:
I think this version clarifies it better. Thanks for your input, it is highly appreciated.

PR #140, opened 2018-05-10
Simplify clone instructions

sorsasampo:
* The github repository `HEAD` already points to `dev`, so that's the branch that will be checked out when cloning.
* `--recurse-submodules` will initialize the submodule(s).


recollir:
LGTM

PR #141, opened 2018-05-10
Add build instructions for single user

sorsasampo:


sorsasampo:
Personally I think it's best to use [stow](https://www.gnu.org/software/stow/):

```
cmake -DCMAKE_BUILD_TYPE=release -DCMAKE_INSTALL_PREFIX=$HOME/stow/timewarrior .
make
make install
cd ~/stow
stow -v timewarrior # symlinks to $HOME/bin etc.
```

Stow can uninstall the symlinks, too.

But I think a `$HOME` prefix example is what most people find useful, as equivalent of:

    ./configure --prefix=$HOME && make &&  make install

`stow` users can modify accordingly once they know `-DCMAKE_INSTALL_PREFIX` is what they want to set.

lauft:
I would mention the `-DCMAKE_INSTALL_PREFIX` but without superimposing any exact path. I would propose to change the build and install code example to something like
---8<---
Build Timewarrior, optionally run the test suite (note: the tarball does not contain tests), and install it.
```
cmake -DCMAKE_BUILD_TYPE=release [-DCMAKE_INSTALL_PREFIX=/path/to/your/install/location]
make
[make test]
[sudo] make install
```
Add the optional parameter `-DCMAKE_INSTALL_PREFIX` if you want to install Timewarrior at a location different from `/usr/local`. The `make install` command may require `sudo` depending on your choice of install location.
--->8---

sorsasampo:
@lauft Just mentioning `-DCMAKE_INSTALL_PREFIX` is okay by me, but instead of shell pseudocode, I would much rather stick to commands that can be copy-pasted. That's why I added a section with alternative commands in the first place.

Anyone else got opinions? :)

lauft:
With the `/path/to/your/install/location` bit, one need to modify the copied string anyway. To keep the code example a little bit more tidier I would move the `-DCMAKE_INSTALL_PREFIX=/path/to/your/install/location` completely down to the explanation paragraph:

---8<---
Build Timewarrior, optionally run the test suite (note: the tarball does not contain tests), and install it.
```
cmake -DCMAKE_BUILD_TYPE=release
make
[make test]
sudo make install
```
Add the optional parameter `-DCMAKE_INSTALL_PREFIX=/path/to/your/install/location` to the `cmake` command if you want to install Timewarrior at a location other than `/usr/local`. The `make install` command may require `sudo` depending on your choice of install location.
--->8---
Preferably IMHO manual installation should be the exception (to installation via package) and for the experienced user - who knows better than just blindly copy-paste ;).

lauft:
I have updated the build instructions as proposed in my last comment. Thanks for your input!

PR #145, opened 2018-05-13
Handling modifications to description/tag/project for task

### LINKED ISSUE #105: [TI-101] Keeps tracking renamed and stopped task
_Antanas B. on 2018-01-11T09:25:17Z says:_ 

 How to reproduce:
```
$ task add dummy
Created task 6.
$ task 6 start
[...]
$ timew
Tracking dummy
Started 2018-01-11T11:18:24
Current 39
Total 0:00:15
$ task 6 mod foo
Modifying task 6 'foo'.
Modified 1 task.
$ task 6 stop
The current interval does not have the 'foo' tag.
Stopping task 4a6c05ca-1396-4a3b-acc5-d04f8afe3a3f 'foo'.
Stopped 1 task.
$ timew
Tracking dummy
[...]
```

vedharish:
The taskwarrior extension script `on-modify.timewarrior` did not forward
modifications for description/tag/project to `timew`.

This resulted in non-existent tasks (with old tag etc) contiued to be
tracked and not updated

[Closes #105](https://github.com/GothenburgBitFactory/timewarrior/issues/105)


ext/on-modify.timewarrior
```diff
@@ -42,6 +42,18 @@ old = json.loads(sys.stdin.readline())
...
+    # Extract attributes for use as tags.
+    tags = [json_obj['description']]
+
+    if 'project' in json_obj:
+        tags.append(json_obj['project'])
+
+    if 'tags' in json_obj:
+        tags.extend(json_obj['tags'])
+
+    return ' '.join(['"%s"' % tag for tag in tags]).encode('utf-8').strip()
```
lauft: Please use new style python formatting.

ext/on-modify.timewarrior
```diff
@@ -53,16 +65,19 @@ elif 'start' not in new and 'start' in old:
...
-    if 'tags' in new:
-        tags.extend(new['tags'])
+# Modifications to task other than start/stop
+elif 'start' in new and 'start' in old:
+    old_combined = get_timew_name_from_json(old)
+    new_combined = get_timew_name_from_json(new)
 
-    combined = ' '.join(['"%s"' % tag for tag in tags]).encode('utf-8').strip()
+    if old_combined != new_combined:
+        command = 'timew stop {0} :yes'.format(old_combined)
```
lauft: Is stopping the old task and starting a new task the right way to react to a modification? When you modify a running task, wouldn't it be better to remove the old tags and add the new ones?

The resulting call to Timewarrior would then be
```
timew untag @1 <old_combined>
timew tag @1 <new_combined>
```

vedharish:
@lauft 

1. Does it make more sense to do `timew tag @1 desc2 && timew untag @1 desc1` instead of doing a stop for the older task and a start later [here](https://github.com/GothenburgBitFactory/timewarrior/pull/145/files#diff-d4c769305f82d6dd14af8c36a756e8c2R79)?

2. If the answer to `1` is `yes`, then what should be the behaviour in this scenario -
```bash
task add desc1
task 1 start
task 1 stop
task 1 mod desc2
```
Should `timew summary` have tag `desc2` here instead of `desc1`

ext/on-modify.timewarrior
```diff
@@ -42,6 +42,18 @@ old = json.loads(sys.stdin.readline())
 new = json.loads(sys.stdin.readline())
 print(json.dumps(new))
 
+def get_timew_name_from_json(json_obj):
```
lauft: I would rename the function to `extract_timew_tags_from_json` or `extract_timew_tags_from` to better reflect its purpose.

ext/on-modify.timewarrior
```diff
@@ -53,16 +65,19 @@ elif 'start' not in new and 'start' in old:
...
 
-    combined = ' '.join(['"%s"' % tag for tag in tags]).encode('utf-8').strip()
+    if old_combined != new_combined:
+        command = 'timew stop {0} :yes'.format(old_combined)
+        subprocess.call(shlex.split(command))
 
-    command = 'timew {0} {1} :yes'.format(start_or_stop, combined)
-    subprocess.call(shlex.split(command))
+        command = 'timew start {0} :yes'.format(new_combined)
+        subprocess.call(shlex.split(command))
```
lauft: One could inline the command string...

lauft:
I would argue that what happens to non-active tasks does not concern Timewarrior. So in your example
```
1) task add desc1
2) task 1 start
3) task 1 stop
4) task 1 mod desc2
```
line 2 would cause `timew start desc1` and line 3 `timew stop desc1`. Line 4) would have no effect on Timewarrior, because the task is not active.

I admit this is tricky. The question to answer here is "What does the user mean when modifying a(n active) task and how should Timewarrior reflect it?" The only general answer to this I can think of is "Please avoid it and expect manual post-processing..."

pbeckingham:
I agree with @lauft, this hook script it intended to capture start/stop info from Taskwarrior only.

As a hook script external to either project, anyone is free to modify and extend the capabilities to meet their needs.  Having implemented and tested this script, they are also free to share and support it.

vedharish:
@lauft @pbeckingham I have made the suggested changes.

PR #157, opened 2018-08-06
CLI: Ensure IDs are non-zero

janikrabe:
This fixes out-of-bounds accesses in several subcommands when the invalid ID `@0` is passed to the program.

### DIFF
--- a/test/ids.t
+++ b/test/ids.t
@@ -50,7 +50,7 @@ class TestIds(TestCase):
 
     def test_should_fail_on_zero_id(self):
         code, out, err = self.t.runError("delete @0")
-        self.assertIn("IDs must be specified.", err)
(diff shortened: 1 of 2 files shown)


test/ids.t
```diff
@@ -48,6 +48,10 @@ class TestIds(TestCase):
         self.assertIn(' @1 ', out)
         self.assertIn(' @2 ', out)
 
+    def test_should_fail_on_zero_id(self):
+        code, out, err = self.t.runError("delete @0")
+        self.assertIn("IDs must be specified.", err)
```
lauft: Could we use an error message like "@0 is not a valid ID"? I think it addresses the problem better as an ID *is* specified, it is just not valid...

janikrabe: Yes, that's a good idea. Will do.

Issue #113, opened 2018-03-07
summary does not display correctly open intervals spanning day border


lauft:
When starting an interval at 23:30 and calling `timew summary` at 0:30 the interval is split at the day change. However, the calculated time is wrong:
```
$ faketime 2018-03-07T0:30 src/timew start 2018-03-06T23:30 
Tracking 
  Started 2018-03-06T23:30:00
  Current         07T00:30:00
  Total               1:00:00
$ faketime 2018-03-07T0:30 src/timew summary 2018-03-06 - 2018-03-08 

Wk  Date       Day Tags    Start End    Time   Total
W10 2018-03-06 Tue      23:30:00   - 1:00:00 1:00:00
W10 2018-03-07 Wed       0:00:00   - 0:30:00 0:30:00
                                                    
                                             1:30:00
```
The first chunk only spans from 23:30 to 0:00 but is accounted with 1:00h.

Closed intervals are displayed correctly:
```
$ faketime 2018-03-07T0:30 src/timew track 2018-03-06T23:30 - 2018-03-07T00:30 
Recorded 
  Started 2018-03-06T23:30:00
  Ended           07T00:30:00
  Total               1:00:00

$ faketime 2018-03-07T0:30 src/timew summary 2018-03-06 - 2018-03-08 

Wk  Date       Day Tags    Start     End    Time   Total
W10 2018-03-06 Tue      23:30:00 0:00:00 0:30:00 0:30:00
W10 2018-03-07 Wed       0:00:00 0:30:00 0:30:00 0:30:00
                                                        
                                                 1:00:00
```

PR #158, opened 2018-08-07
CmdSummary: Do not extend end time beyond end of day

janikrabe:
When showing a summary, set the end time to the current time on the last day only. This prevents over-counting of time when a tracked interval covers multiple days.

Fixes test `TestSummary.test_with_open_interval` when the system time is between midnight and 01:00 AM.


lauft:
LGTM, thanks for your contribution!

Closes #113

PR #162, opened 2018-08-08
Database: Fix an incorrect comment

janikrabe:


src/Database.cpp
```diff
@@ -310,16 +310,16 @@ unsigned int Database::getDatafile (int year, int month)
...
 //
 //   2016-02-20 to 2016-04-15
 //
-// Given the monthly storage scheme, split the Datarange into a vector of
-// segmented Dataranges:
+// Given the monthly storage scheme, split the Daterange into a vector of
+// segmented Dateranges:
 //
-//   2016-02-20 to 2016-03-01
+//   2016-02-01 to 2016-03-01
```
lauft: Should be 2016-02-20, as it is the start date of the example above

src/Database.cpp
```diff
@@ -310,16 +310,16 @@ unsigned int Database::getDatafile (int year, int month)
...
 //   2016-03-01 to 2016-04-01
-//   2016-04-01 to 2016-05-15
+//   2016-04-01 to 2016-05-01
```
lauft: Should be 2016-04-15, as it is the end date of the example above. Here, the month is actually wrong, not the day

janikrabe:
@lauft [`segmentRange()`](https://github.com/GothenburgBitFactory/timewarrior/blob/1.1.1/src/Database.cpp#L271) returns the database segments (full months) covering a given range. Both [`segmentStart`](https://github.com/GothenburgBitFactory/timewarrior/blob/1.1.1/src/Database.cpp#L289) and [`segmentEnd`](https://github.com/GothenburgBitFactory/timewarrior/blob/1.1.1/src/Database.cpp#L300) are always on the first day of a month.

Adding `std::cout << "--> " << Range (segmentStart, segmentEnd).dump() << std::endl;` before [`L301`](https://github.com/GothenburgBitFactory/timewarrior/blob/1.1.1/src/Database.cpp#L301) yields:

```text
$ src/timew track 2016-02-20 - 2016-04-15
--> Range 2016-02-01T00:00:00 - 2016-03-01T00:00:00
--> Range 2016-03-01T00:00:00 - 2016-04-01T00:00:00
--> Range 2016-04-01T00:00:00 - 2016-05-01T00:00:00
```

lauft:
Good read, you are right! I am taking all back :)

PR #163, opened 2018-08-08
Add *.pyc files to .gitignore

janikrabe:
Running 'make test' generates 'ext/totals.pyc', which should not be under version control.


lauft:
OK.

PR #170, opened 2018-08-26
Fix getTracked() when last interval is empty

### LINKED ISSUE #166: getTracked behaves incorrectly when latest interval is empty
```text
$ timew start
  Tracking […]

$ timew stop
  Recorded […]

$ timew resize @1 0
  Resized @1 to 0:00:00

$ timew start
  Tracking  […]

$ timew stop
  Recorded  […]

$ timew resize @1 0
  Resized @1 to 0:00:00

$ timew summary :year :ids
  (shows both IDs, @1 and @2, as expected)

$ timew move @2 2018-01-01
  ID '@2' does not correspond to any tracking.
```

This is due to how `getTracked()` computes the default filter if no other filter is specified: it uses the earliest start time and the latest end time of any interval in the database. This results in an interval like this:

```text
filter:  [----------)
     A:  Z
     B:             Z
```

Here, `B` is a zero-width interval, so it does not overlap `filter` and is excluded from the list of tracked intervals returned by `getTracked()`.

This also means that attempting to modify `@n` will instead modify `@n+1`.

janikrabe:
Fixes #166
Depends on #169

### DIFF
--- a/src/data.cpp
+++ b/src/data.cpp
@@ -595,9 +595,9 @@ std::vector <Interval> getTracked (
   auto inclusions = getAllInclusions (database);
 
   // Exclusions are only usable within a range, so if no filter range exists,
-  // determine the outermost range of the inclusions, ie:
+  // determine the infinite range starting at the first inclusion, i.e.:
   //
-  //   [earliest start, latest end)
+  //   [earliest start, infinity)
   //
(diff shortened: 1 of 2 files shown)


lauft:
I have merged your pull request manually. Thank you for your contribution!

Issue #159, opened 2018-08-07
Tags enclosed in escaped double-quotes are not handled correctly


ghost:
For a reason yet unknown, I cannot stop tracking time:
```
$ timew :debug                                                                                                                                                                                                                                                        255 
CLI Parser
  _original_args
    timew :debug
  _args
    word basename='timew' raw='timew' BINARY 
    word canonical=':debug' raw=':debug' ORIGINAL HINT FILTER 

>> 2017-11.data: 24 intervals
>> 2018-04.data: 32 intervals
>> 2018-05.data: 67 intervals
>> 2018-06.data: 78 intervals
>> 2018-07.data: 10 intervals
>> 2018-08.data: 10 intervals
Tracking "Task#2 - ... .. https://..." project:x
  Started 2018-08-07T10:15:00
  Current            13:12:36
  Total               6:59:38
>> Timer timew 0.012903 sec

$ timew stop :debug
CLI Parser
  _original_args
    timew stop :debug
  _args
    word basename='timew' raw='timew' BINARY 
    word canonical='stop' raw='stop' ORIGINAL CMD 
    word canonical=':debug' raw=':debug' ORIGINAL HINT FILTER 

>> 2017-11.data: 24 intervals
>> 2018-04.data: 32 intervals
>> 2018-05.data: 67 intervals
>> 2018-06.data: 78 intervals
>> 2018-07.data: 10 intervals
>> 2018-08.data: 10 intervals
>> Found 16 holidays
>> Found 7 exclusions
>> Loaded 221 tracked intervals
>> Input         interval 2018-08-07T10:15:00 - 2018-08-07T13:13:09 # "Task#2 - ... .. https://..." project:x
>> Overlaps with
>>               interval @1 2018-08-07T10:15:00 # "Task#2 - ... .. https://..." project:x
You cannot overlap intervals. Correct the start/end time, or specify the :adjust hint.
>> Timer timew 0.073525 sec
```

The interval timewarrior sees the overlap with appears to the exact same interval I want to close using the `timew stop` command.

```
$ timew diagnostics :debug                                                                                                                                                                                                                                            255 
CLI Parser
  _original_args
    timew diagnostics :debug
  _args
    word basename='timew' raw='timew' BINARY 
    word canonical='diagnostics' raw='diagnostics' ORIGINAL CMD 
    word canonical=':debug' raw=':debug' ORIGINAL HINT FILTER 


timew 1.1.1
       Platform: Linux

[...]

PR #173, opened 2018-09-01
Interval: prevent adding empty tags to Interval

taiyu-len:
Fix bug where having an empty tag in an Interval inserts an extra space
into its seriliazed string, causing other operations to be unable to
find the interval and stop/remove/modify/etc it

example
```
$ timew start tag "$empty"
# creates entry
# inc 20180831T234143Z #  tag
#                       ^^ extra space gets inserted
$ timew stop
You cannot overlap intervals. Correct the start/end time, or specify the :adjust hint.
# deleteInterval searchs for a line without the extra space, and becomes unable to remove tag
```

an alternative solution is to escape empty strings (inserting `""` as a tag),
but i think its reasonable to assume an empty tag was not intended, perhaps that could be an option.

### DIFF
--- a/src/Interval.cpp
+++ b/src/Interval.cpp
@@ -73,7 +73,7 @@ void Interval::initialize (const std::string& line)
     {
       // Optional <tag> ...
       for (unsigned int i = 2 + offset; i < tokens.size (); ++i)
-        _tags.insert (tokens[i]);
+        tag (tokens[i]);
     }
 
     return;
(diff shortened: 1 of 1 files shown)


lauft:
Good catch! However entering `timew tag "" @1` still returns 
```
timew tag @1 ""
Added  to @1
```
I think an error message like "tags must not be empty" should be returned in this case. 
I would be also good if you could provide tests for this.

lauft:
The solution to always quote tags also came into my mind when working on #159 - it may be a path for the future...

sruffell:
Just a note that commit c1f3d8be72e2f17547, which is in 1.3.0, will prevent the database from being corrupted in this case as well.

Issue #131, opened 2018-04-18
[feature] allow to adjust historical records to a fixed date/time


svenXY:
Hi,

it would be nice if one could adjust the start or end time of a historical record to a fixed date/time, like e.g.

```bash
timew ended @1 23:00
timew started @1 11:30
```

Very often, I can more or less remember when I switched to a different task, but then having to calculate how much I have to lengthen or shorten or whatever is very time-consuming

lauft:
I like the idea, but I would alter the command to 
```
timew modify (start|end) <id> <date>
```
This is more in the style "TimeWarrior do <foo>" like the other commands - plus one could integrate shorten and lengthen
```
timew modify length <id> (|+|-)<duration>
```

svenXY:
sounds good to me.

sruffell:
I also like @lauft's proposed syntax. I wonder if anyone does not like it?  I hit this enough that I probably should write my own pull request to add this feature.

sruffell:
Just a heads up, that I've also pushed a branch to my fork of timewarrior that has this change applied against the 1.1.1 branch so I can start using it without modifying the format of the database.  When I tried to run with the branch based off dev, I ran into problems since the tags database wasn't updated and there isn't a 'standard' way to migrate yet, like as proposed in #182.

So anyone who is interested can use the commit at the tip of the branch below with their current database:
https://github.com/sruffell/timewarrior/tree/modify_cmd_1.1.1

sorsasampo:
@lauft anything stopping @sruffell's commits from being merged?

I think this feature is important enough to warrant a new release by itself. It's one of the things I miss most from [samg/timetrap](https://github.com/samg/timetrap).

PR #185, opened 2018-11-13
#131: Introduce 'modify' to adjust historical records to fixed date/time.

sruffell:


test/modify.t
```diff
@@ -0,0 +1,158 @@
...
+
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=3)))
+        self.t("stop {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=1)))
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(minutes=49)))
+        self.t("stop")
+
+        code, out, err = self.t("modify stop @2 {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=2)))
+
+        # Verify that the total time is one hour less than it was before.
+        code, out, err = self.t("summary")
```
lauft: Better use `self.t.export()` than parsing the output of `timew summary`, see e.g. `move.t:117`:
```
j = self.t.export()

self.assertEquals(len(j), 2)
self.assertClosedInterval(j[0],
                          expectedStart="20170301T110000Z",
                          expectedEnd="20170301T133000Z",
                          expectedTags=["foo"])
```

test/modify.t
```diff
@@ -0,0 +1,158 @@
...
+        now_utc = datetime.now().utcnow()
+
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=3)))
+        self.t("stop {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=1)))
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(minutes=49)))
+        self.t("stop")
+
+        code, out, err = self.t("modify start @2 {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=2)))
+
+        code, out, err = self.t("summary")
```
lauft: See comment at line 86

test/modify.t
```diff
@@ -0,0 +1,158 @@
...
+        now_utc = datetime.now().utcnow()
+
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=3)))
+        self.t("stop {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=1)))
+        self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(minutes=49)))
+        self.t("stop")
+
+        code, out, err = self.t.runError("modify start @1 {:%Y-%m-%dT%H:%M:%S}Z".format(now_utc - timedelta(hours=2)))
+        self.assertIn("You cannot overlap intervals", err)
+
```
lauft: Test `modify should not move start beyond end`?
Test `modify should not move end before start`?

src/commands/CmdModify.cpp
```diff
@@ -0,0 +1,95 @@
...
+    throw format ("ID '@{1}' does not correspond to any tracking.", id);
+
+  Interval interval = tracked.at (tracked.size () - id);
+  if (interval.is_open ())
+    throw format ("Cannot modify open interval @{1}.", id);
+
+  if (filter.start.toEpoch() == 0)
+    throw std::string ("No updated time specified. See 'timew help modify'.");
+
+  database.startTransaction ();
```
lauft: Maybe move this line down to line 85? Start transaction right before the database stuff happens...

src/commands/CmdHelp.cpp
```diff
@@ -52,6 +52,7 @@ int CmdHelpUsage (const Extensions& extensions)
             << "       timew help [<command> | interval | hints | date | duration]\n"
             << "       timew join @<id> @<id>\n"
             << "       timew lengthen @<id> [@<id> ...] <duration>\n"
+            << "       timew modify (start|stop) @<id> <date>\n"
```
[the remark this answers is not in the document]
sruffell: Ack. I was thinking from a user standpoint I want to modify the argument passed to the `stop` command and that end was internal usage, but I see it referenced in the user documentation.

src/commands/CmdHelp.cpp
```diff
@@ -693,6 +694,26 @@ int CmdHelp (
...
+                << "interval IDs. Using the right ID, you can identify an interval to modify. For\n"
+                << "example, show the IDs:\n"
+                << '\n'
+                << "  $ timew summary :week :ids\n"
+                << '\n'
+                << "Then having selected '@2' as the interval you wish to modify:\n"
+                << '\n'
+                << "  $ timew modify stop @2 2018-10-10T17:10:00\n"
+                << '\n'
+                << "See also 'summary', 'lengthen', and 'shorten'.\n"
```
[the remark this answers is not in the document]
sruffell: Heh, I was not aware of the resize command. I see it in the man pages (which I also see that I did not update for the modify command) but it is not in the output of help.

But in general, ack. I will make this change in the updated pull request.

test/modify.t
sruffell: Ack

test/modify.t
sruffell: Ack

test/modify.t
sruffell: I believe these case are already covered?  I've updated the test description to be of this form in the updated pull request.

src/commands/CmdModify.cpp
sruffell: ack

sruffell:
Thanks for review. I've updated the branch with what I hope are the fixes to all your comments.

Additionally I've:

- Updated the tests to make sure only the start of an open interval can be modified by the end cannot.
- Updated the man page for the `modify` command.

With regards to the man pages, I was not sure of the proper formatting for the non-optional argument to the modify command. I also did not see how the individual command pages get run through the configure with the cmake files. So there is a good chance I missed something there.

src/commands/CmdHelp.cpp
```diff
@@ -693,6 +694,26 @@ int CmdHelp (
                 << "See also 'summary', 'tag', 'untag', 'shorten'.\n"
                 << '\n';
 
+    // Ruler                 1         2         3         4         5         6         7         8
+    //              12345678901234567890123456789012345678901234567890123456789012345678901234567890
+    else if (words[0] == "modify")
+      std::cout << '\n'
+                << "Syntax: timew modify (start|stop) @<id> @<date>\n"
```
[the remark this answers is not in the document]
sruffell: Should be fixed now.

src/commands/CmdModify.cpp
```diff
@@ -0,0 +1,95 @@
...
+  Interval empty_filter = Interval();
+  auto tracked = getTracked (database, rules, empty_filter);
+
+  int id = *ids.begin();
+  if (id > static_cast <int> (tracked.size ()))
+    throw format ("ID '@{1}' does not correspond to any tracking.", id);
+
+  Interval interval = tracked.at (tracked.size () - id);
+  if (interval.is_open ())
+    throw format ("Cannot modify open interval @{1}.", id);
```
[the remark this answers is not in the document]
sruffell: Fixed.

lauft:
Closes #131

PR #200, opened 2018-12-30
Fix on-modify.timewarrior hook for tasks with multi line description

devurandom:
If the description contained multiple lines, it would be formatted as a
bytes object, including Python's `b""` marker for byte literals and any
newlines.  This would then be passed literally to `timew`, which would
then choke on it, because it would record the newlines in its database.

This patch simply gets rid of the string join/split operations and the
encoding, which solves the issue.

Since we pass arrays instead of strings to `subprocess.call`, we are not
subject to command injection security vulnerabilities.

Fixes: 0b6dbf7e12d30fc791a524501e3483989a092a8c

### DIFF
--- a/ext/on-modify.timewarrior
+++ b/ext/on-modify.timewarrior
@@ -28,7 +28,6 @@
 import sys
 import json
 import subprocess
-import shlex
 
 # Hook should extract all of the following for use as Timewarrior tags:
 #   UUID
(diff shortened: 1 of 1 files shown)


devurandom:
> Could you provide an example of the passed JSON data?

I created it, following an [official example](https://taskwarrior.org/docs/examples.html), like this:
```
task add "test
1
2
3"
```

Please note that the current code works fine with Python 2.7, but breaks with Python 3.7. This PR makes it work with both versions.

fdcds:
This also fixes #220 (for which #234 is a different fix) and a Python 3 incompatibility (for which #233 is a different fix), while being way simpler and potentially safer.

PR #202, opened 2019-01-03
Add Travis-based CI service

tbabej:
In Taskwarrior's image, this PR implements a Travis-based continuous integration service across various Linux distros and Mac OS X.

Note that I had to remove (stale?) test/googletest submodule directory (it was not referenced in the .gitmodules, and hance Travis couldn't clone it with ` git submodule update --init --recursive` - see https://travis-ci.org/tbabej/timewarrior/jobs/474635224 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,49 @@
+sudo: required
+language: generic
+matrix:
+  include:
+    - name: "Gentoo (latest)"
+      os: linux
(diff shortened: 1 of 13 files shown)


lauft:
Thanks for setting it up.

Those submodules were added accidentially, so removing them is perfectly fine.

Issue #198, opened 2018-12-26
Problem with `%` symbol in tag


AlexLov:
It's possible to add time interval with `%` symbol in tags but then impossible to sort it and even delete it.

Examples:
```
$ timew
There is no active time tracking.
$ timew start 'ta%g'
Tracking ta%g
  Started 2018-12-26T18:33:58
  Current                  58
  Total               0:00:00
$ timew sum :ids

Wk  Date       Day ID Tags        Start End    Time   Total
W52 2018-12-26 Wed @1 %, g, ta 18:33:58   - 0:00:07 0:00:07

                                                    0:00:07

$ timew stop :debug
CLI Parser
  _original_args
    timew stop :debug
  _args
    word basename='timew' raw='timew' BINARY
    word canonical='stop' raw='stop' ORIGINAL CMD
    word canonical=':debug' raw=':debug' ORIGINAL HINT FILTER

>> 2018-12.data: 1 intervals
>> Found 0 holidays
>> Found 0 exclusions
>> Loaded 1 tracked intervals
>> Input         interval 2018-12-26T18:33:58 - 2018-12-26T18:40:29 # % g ta
>> Overlaps with
>>               interval @1 2018-12-26T18:33:58 # % g ta
You cannot overlap intervals. Correct the start/end time, or specify the :adjust hint.
>> Timer timew 0.011393 sec

$ timew delete @1 :debug
CLI Parser
  _original_args
    timew delete @1 :debug
  _args
    word basename='timew' raw='timew' BINARY
    word canonical='delete' raw='delete' ORIGINAL CMD
    word raw='@1' value='1' ORIGINAL ID FILTER
    word canonical=':debug' raw=':debug' ORIGINAL HINT FILTER

>> 2018-12.data: 1 intervals
>> Found 0 holidays
>> Found 0 exclusions
>> Loaded 1 tracked intervals
Deleted @1
>> Timer timew 0.010465 sec

$ timew sum :ids

Wk  Date       Day ID Tags        Start End    Time   Total
W52 2018-12-26 Wed @1 %, g, ta 18:33:58   - 0:00:19 0:00:19

                                                    0:00:19

$
```

```
$ timew diagnos

timew 1.1.1
       Platform: Darwin

Compiler:
        Version: 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.10.43.1)
[...]

PR #203, opened 2019-01-04
Fix segfaults on join command and invalid tags database.

sruffell:
Two commits that eliminate segment faults when attempting to join intervals and when the tags database is missing "count" member for a particular tag name.

### DIFF
--- a/src/Database.cpp
+++ b/src/Database.cpp
@@ -372,8 +372,10 @@ void Database::initializeTagDatabase ()
   {
     auto key = pair.first;
     auto* value = (json::object*) pair.second;
-    auto* number = (json::number*) value->_data["count"];
-
+    auto iter = value->_data.find ("count");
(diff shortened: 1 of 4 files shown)


sruffell:
Hmm...my branch name is bad. It should be for 2_0_0.

lauft:
1.2.0 is a remnant of the old practice to start a new branch with the next upcoming release. It will be fast-forwarded to dev as soon as 1.2.0 is released and/or work on 2.0.0 is started

sruffell:
Thanks. Is there anything else I should do with this pull request then? It's still set to merge into dev, but the name is odd. Should I fix the name and resubmit or just leave it as is?

lauft:
I do not see it necessary to rename it. Code looks good to me and it fixes #198 - so I will rebase and merge it.

PR #208, opened 2019-02-17
Do not assume /\A["']/ strings are quoted

janikrabe:
Do not assume that strings starting with a quote are quoted. For instance, the string `"a b` contains a space that needs to be escaped, so quoteIfNeeded should return `"\"a b"` rather than `"a b`.

Found while debugging GothenburgBitFactory/libshared#14.


lauft:
With the following example
```
#!/bin/bash

export TIMEWARRIORDB=/tmp/tiwmewarriordb
mkdir -p ${TIMEWARRIORDB}/data
rm -f ${TIMEWARRIORDB}/data/*.data
:> ${TIMEWARRIORDB}/timewarrior.cfg

src/timew start '"a'
src/timew stop
```
I get
```
Note: '"a' is a new tag.
Tracking "a
  Started 2019-03-03T15:48:36
  Current                  36
  Total               0:00:00
Trying to decrement non-existent tag '"a'
```
before and 
```
Note: '"\"a"' is a new tag.
Tracking "\"a"
  Started 2019-03-03T15:47:52
  Current                  52
  Total               0:00:00
Trying to decrement non-existent tag '"a'
```
after applying your PR. I will have to investigate this before merging your PR.

janikrabe:
For future reference, these messages are from `src/Database.cpp:113` and `src/helper.cpp:111`. Both of these call `quoteIfNeeded`. It might make sense to either remove the `quoteIfNeeded` for display purposes, or to remove the additional surrounding single quotes from the "new tag" message.

The issue with the non-existent tag is GothenburgBitFactory/libshared#14.

Issue #212, opened 2019-02-26
--help should show help, not considered to be a tag


varac:
I ran into this multiple times. Please show help when `--help` is used, and not consider it a tag.
Most cli tools respect this flag and show their help/usage.
It's especially helpful for new users.

```
$ timew track --help
Note: '"--help"' is a new tag.
Tracking "--help"
  Started 2019-02-26T08:18:35
  Current                  35
  Total               0:00:00
```

lauft:
I agree. Timewarrior should process `--help` as expected.

lauft:
Closed by #219

PR #219, opened 2019-04-07
#212: "--help" should show help, not considered to be a tag.

sruffell:
This change makes any instance of '--help' on the command line to be made the
command for this particular invocation.

Also note, this isn't something I hit very often but just wanted to take the opportunity to use the issue to learn another part of the code since I use timewarrrior every day.

### DIFF
--- a/src/CLI.cpp
+++ b/src/CLI.cpp
@@ -445,6 +445,23 @@ void CLI::canonicalizeNames ()
       alreadyFoundCmd = true;
     }
 
+    // 'timew <command> --help' should be treated the same as 'timew help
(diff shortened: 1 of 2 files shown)


sruffell:
It is not clear to me if the checks that have failed are problems with the checks or with my proposed change.

lauft:
I think those failed because of infrastructure problems: the build for gentoo seems to stall at the setup phase and the ubuntu build broke because a test assumed it would run before 11:59pm (and somehow it did not).

lauft:
Thanks for your contribution!

lauft:
I am going to extend this even one step further and will include `-h` as well

PR #227, opened 2019-05-29
MAINT: on-modify.timewarrior imports from future

### LINKED ISSUE #216: on-modify.timewarrior assumes python is python 3
it uses print() and there's no

    from __future__ import print_function

so I suppose it won't work on systems where python points to python 2?

I know the trend is to deprecate python 2 wherever possible, but this goes back to version 1.1.1 of timewarior, when there were slightly less reasons to not support python 2

Is it possible to make this hook python 2 compatible (especially that the more recent ext/totals.py does it by making use of the aforementioned import from future voodoo)?

quazgar:
This should fix #216.

Untested commit.

### DIFF
--- a/ext/on-modify.timewarrior
+++ b/ext/on-modify.timewarrior
@@ -30,7 +30,6 @@ import json
 import subprocess
 import shlex
 
-# For issue #216 and generally sounds like a good idea
 from __future__ import print_function
 
 # Hook should extract all of the following for use as Timewarrior tags:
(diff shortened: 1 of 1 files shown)


ext/on-modify.timewarrior
```diff
@@ -30,6 +30,9 @@ import json
 import subprocess
 import shlex
 
+# For issue #216 and generally sounds like a good idea
```
lauft: Please remove this comment (If you like, put it into the commit message)

quazgar: Removed.

PR #233, opened 2019-07-10
on-modify.timewarrior: Fix timewarrior tags being encoded as byte arrays instead of strings

fdcds:
This prevents an issue where timewarrior would be `Tracking "b\"...\""` with Python 3.6.8


fdcds:
#200 is a more correct and powerful version of this.

Issue #220, opened 2019-04-15
Tag containing double quotes breaks TimeWarrior v1.2.0


sskras:
And if I use double quotes from the Task Warrior, TimeW doesn't break.

I tested TimeWarrior in sandbox on MSys2, not on Linux:
```
$ timew --version
1.2.0

$ uname -a
MSYS_NT-10.0 SAUKRSXM1 2.11.1(0.329/5/3) 2018-09-10 14:19 x86_64 Msys

$ rm -r $TIMEWARRIORDB
$ rm -r $TASKDATA

$ export TASKDATA=/tmp/.task
$ export TIMEWARRIORDB=/tmp/.timew

$ mkdir -p $TASKDATA/hooks
$ mkdir $TIMEWARRIORDB

# Both showing empty lists:

$ task
TASKDATA override: /tmp/.task
[task next]
No matches.

$ timew summ
No filtered data found in the range 2019-04-15T00:00:00 - 2019-04-16T00:00:00.

# Link them together:

$ cp -v /usr/local/share/doc/timew/ext/on-modify.timewarrior /tmp/.task/hooks/
'/usr/local/share/doc/timew/ext/on-modify.timewarrior' -> '/tmp/.task/hooks/on-modify.timewarrior'
```

Initialize two tasks:

```
# Task #1 created:

$ task add proj:TEST 'first task'
TASKDATA override: /tmp/.task
Created task 1.
The project 'TEST' has changed.  Project 'TEST' is 0% complete (1 task remaining).

$ task
TASKDATA override: /tmp/.task
[task next]

ID Age Project Description Urg
 1 3s  TEST    first task     1

1 task

# ... and started:

$ task 'first task' start
Starting task 228da0ea-13e0-444e-922e-dcc326a175ea 'first task'.
Started 1 task.
TASKDATA override: /tmp/.task
Note: 'TEST' is a new tag.
Note: '"first task"' is a new tag.
Tracking TEST "first task"
  Started 2019-04-15T18:50:02
  Current                  02
  Total               0:00:00
Project 'TEST' is 0% complete (1 task remaining).

# ... it's being time-tracked too:

$ timew summ

Wk  Date       Day Tags                Start End    Time   Total
W16 2019-04-15 Mon TEST, first task 18:50:02   - 0:00:05 0:00:05

                                                         0:00:05

# Task #2 created (including double quotes):

$ task add proj:TEST 'second task containing \"double\" quotes'
[...]

PR #234, opened 2019-07-10
on-modify.timewarrior: Quote arguments to `timew` properly (#220)

fdcds:
Closes: #220


fdcds:
#200 is a more correct and powerful version of this.

PR #235, opened 2019-07-10
on-modify.timewarrior: Use JIRA task ID as a tag, if present

fdcds:
If the task was imported from JIRA using bugwarrior, use the JIRA task ID as an additional tag.

### DIFF
--- a/ext/on-modify.timewarrior
+++ b/ext/on-modify.timewarrior
@@ -48,6 +48,9 @@ def extract_timew_tags_from(json_obj):
     # Extract attributes for use as tags.
     tags = [json_obj['description']]
 
+    if 'jiraid' in json_obj:
+        tags.append(json_obj['jiraid'])
+        
     if 'project' in json_obj:
         tags.append(json_obj['project'])
 


lauft:
As far as I understand https://taskwarrior.org/docs/design/task.html#atts, `jiraid` is not a standard attribute of Taskwarrior but entered by Bugzilla. How is this handled for other bug-trackers (of whom Bugzilla [supports a lot](https://bugwarrior.readthedocs.io/en/latest/services.html))?

I would argue it is a bug of Bugzilla not to add the JIRA-ID as a tag to the task...

The hook script should not depend on/implement other project's features than Taskwarrior's.

fdcds:
> I would argue it is a bug of Bugzilla not to add the JIRA-ID as a tag to the task...

Well, bugwarrior adds the key:value to the task, which can be queried like tags. So from a taskwarrior point of view having an additional tag would be redundant. Only for timewarrior these key:value attributes need to be translated into tags.

> The hook script should not depend on/implement other project's features than Taskwarrior's.

I understand your POV, but that would mean that the `~/.task/hooks/on-modify.timewarrior` needs to call additional `~/.timewarrior/hooks/on-modify.bugwarrior` hooks to turn other project's taskwarrior task attributes into timewarrior tags. This seems very inconvenient in several aspects, one being more hooks and infrastructure to support them (making the whole thing more prone to issues, bugs, misconfigurations), but more importantly the data is only available in taskwarrior, so `~/.task/hooks/on-modify.*` is the only place where we can get it without jumping through nasty hoops.

fdcds:
> How is this handled for other bug-trackers (of whom Bugzilla supports a lot)?

They would need to be added to the hook individually.

> I would argue it is a bug of Bugzilla not to add the JIRA-ID as a tag to the task...

Well, bugwarrior adds `key:value` attributes to the task, which can be queried like tags. So from a taskwarrior point of view having an additional tag would be redundant. Only for timewarrior these `key:value` attributes need to be translated into tags.

> The hook script should not depend on/implement other project's features than Taskwarrior's.

I understand your POV, but that would mean that the `~/.task/hooks/on-modify.timewarrior` needs to call additional `~/.timewarrior/hooks/extract-tags-from-taskwarrior.bugwarrior` hooks to turn other project's taskwarrior task attributes into timewarrior tags (`~/.task/hooks/on-modify.*` is the only place where we can get it without jumping through nasty hoops). Doable, but makes `on-modify.timewarrior` more complex.

lauft:
I agree that this is a simple solution, but it feels like opening Pandora's box by allowing `jiraid` in and then having all the other services come knocking at the door...

I wonder whether an (user-defined-)attribute-2-tag-map could solve this situation (in the future).

fdcds:
> I wonder whether an (user-defined-)attribute-2-tag-map could solve this situation (in the future).

I like the idea, but it would have to be a more general function. For JIRA I can take the `jiraid`, which includes the project, to create a unique ID tag. But bugwarrior's GitHub support, for example, sets `githubnumber` and `githubrepo` attributes, which one would want to combine into `githubrepo#githubnumber` to produce a unique ID tag. My suggestion would be to pass the complete taskwarrior task JSON object to the function and have it produce a list of timewarrior tag strings.

xerus2000:
yeah, this needs a more generic solution, supporting a single custom property would be weird

bserem:
In bugwarrior github, we can add the issue-id as a tag in **taskwarrior** using this inside bugwarrior:
`github.add_tags = {{githubnumber}}`

Similarly, I would expect that this would work:
`jira.add_tags = {{jiraid}}`

This would add the tag in the task, so the list of tags would already be ok when the hook is called.
You would need to prepend it with something so that timewarrior will get triggered, like:
`github.add_tags = #{{githubnumber}}`
`jira.add_tags = #{{jiraid}}`
or use the approach in #425 

So, my suggestion is that this can be handled in bugwarrior right away actually.

lauft:
Migrated to https://github.com/GothenburgBitFactory/task-timewarrior-hook/pull/3

Issue #41, opened 2018-02-11
[TI-36] Hidden and/or non-sorting tags


taskwarrior:
_Paul Kishimoto on 2016-08-19T21:17:02Z says:_ 

 Also related to [this extension|https://gist.github.com/khaeru/860b60a4e42420fe35f70b7a98294f6e], Toggl reports a unique ID for every entry. At first, I had the extension write these as a tag like `toggl_id:ID`. But I found this prevented Timewarrior from grouping these intervals together for reporting.

I also use a variation on the [`on-modify.timewarrior` hook|https://git.tasktools.org/projects/TM/repos/timew/browse/ext/on-modify.timewarrior] for Taskwarrior that adds a tag like `uuid:<UUID>` to Timewarrior intervals. In this case, the UUID is useful for later matching, especially if the task description is changed in Taskwarrior; but it would be useful to be able to _omit_ it from Timewarrior displays.

Perhaps:
*  Tags beginning with `_`, `.` or some other special character could be hidden and/or ignored when reporting, and/or
*  A configuration option like `tag.<tag>.ignore = yes` could be added.

taskwarrior:
Migrated metadata:
```
Created: 2016-08-19T21:17:02Z
Modified: 2018-01-29T19:53:04Z
```

taskwarrior:
_Paul Beckingham on 2017-01-17T02:16:10Z says:_ 

 I agree, with a tags-only interface, hiding tags is useful.

taskwarrior:
_Anthony CARRÉ on 2018-01-29T19:53:04Z says:_ 

 hidden uid tag could probably help for renamed tasks time tracking

danihodovic:
Adding task uuid's to timewarrior data is useful for correlating intervals with tasks in third party programs.

xerus2000:
Yes, this is a good idea!

I also see the fact that the timewarrior description is added as a normal tag by the hook problematic: If you adjust the task, you can't really trace that back. Also that means that in the tags overview there will be one tag for every task you ever tracked...

PR #236, opened 2019-07-10
on-modify.timewarrior: Use last annotation as additional tag, if present

fdcds:
This allows the following workflow to be effective and tracked in a useful manner:
```
task 123 start First draft
task 123 stop
...
task 123 start Second iteration
task 123 stop
...
task 123 start Bugfixing
task 123 stop
...
```

This will be tracked as:
```
01:00 - 02:00 <TASK 123 TAGS> "First draft"
03:00 - 04:00 <TASK 123 TAGS> "Second iteration"
05:00 - 06:00 <TASK 123 TAGS> "Bugfixing"
```

### DIFF
--- a/ext/on-modify.timewarrior
+++ b/ext/on-modify.timewarrior
@@ -54,6 +54,9 @@ def extract_timew_tags_from(json_obj):
     if 'tags' in json_obj:
         tags.extend(json_obj['tags'])
 
+    if 'annotations' in json_obj:
+        tags.append(json_obj['annotations'][-1]['description'])
+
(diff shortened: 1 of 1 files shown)


xerus2000:
> The problem I see here is that this creates a lot of new tags (which are only used once).
 reports

As @JavaZauber mentioned, this happens already with descriptions, so the argument is hardly valid.
We need to solve #41/#209/#179 anyways, and then this won't be a problem anymore either.

lauft:
Migrated to https://github.com/GothenburgBitFactory/task-timewarrior-hook/pull/4

PR #241, opened 2019-08-02
Allow continue by tag.

croesch:
Hi,

I'm using timewarrior in a way where I switch a lot between small tasks and most of the time remember one of the tags of an interval I'd like to continue. Currently I do something like:

```
$ timew summary :month :ids XYZ
$ timew continue @LAST_ID_FOR_XYZ
```

I'm not sure if I'm the only one so I had a look and found it quite possible to introduce the desired behaviour. Is that code anywhere near acceptable or does it fail a lot of your checks? Please let me know what needs to be improved.

I'd like to see that included in the next timew release. (btw is that planned?)

Do you need an issue for that or is the merge request sufficient? If you'd accept the behavior I can surely adopt the documentation, too.

### DIFF
--- a/src/commands/CmdContinue.cpp
+++ b/src/commands/CmdContinue.cpp
@@ -38,7 +38,7 @@ int CmdContinue (
 {
   // Gather IDs and TAGs.
   std::set <int> ids = cli.getIds();
-  auto tags = cli.getTags();
+  auto filter = getFilter (cli);
 
   if (ids.size() > 1)
     throw std::string ("You can only specify one ID to continue.");
@@ -50,8 +50,8 @@ int CmdContinue (
   {
     // Load the data.
     // Note: There is no filter.
-    Interval filter;
-    auto tracked = getTracked (database, rules, filter);
+    Interval noFilter;
+    auto tracked = getTracked (database, rules, noFilter);
 
     auto id = *ids.begin ();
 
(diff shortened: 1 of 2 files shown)


lauft:
This sounds like a good idea.

However, as I am trying to finalize 1.2.0, I do not want to include last minute features. So this will go into the following release.

src/commands/CmdContinue.cpp
```diff
@@ -59,6 +60,34 @@ int CmdContinue (
 
     to_copy = tracked[tracked.size () - id];
   }
+  else if (!tags.empty())
```
lauft: Do not implement tag filtering here, use the available on in `getTracked` (You need to create a filter with the given tag set). With that you should only need to pick the first interval from the returned array

test/continue.t
```diff
@@ -109,6 +109,58 @@ class TestContinue(TestCase):
...
+        self.assertIn("Tracking FOO\n", out)
+
+        code, out, err = self.t("start BAR 30min ago")
+        self.assertIn("Tracking BAR\n", out)
+
+        code, out, err = self.t("continue FOO")
+        self.assertIn("Recorded BAR\n", out)
+        self.assertIn("Tracking FOO\n", out)
+
     def test_continue_with_id_and_date(self):
```
lauft: I am missing a test to exclude continue by tag *and* id

croesch:
I'm not quite sure if you meant it like that regarding the filter, but I added two tests (one I missed and one you missed).

Regarding specifying an id and tags I have no idea which behavior would be expected if you type an id and tags that don't match. What do you think?

lauft:
I would define the behaviour of `continue by tag` as follows:

`timew continue TAG... [datetime]` continues the latest interval which contains all the tags given in its tag set at the given datetime.

Given the following intervals and their ids
```
20190810T130000 - 20190810T140000 BAR         [@4]
20190810T140000 - 20190810T150000 FOO BAR     [@3]
20190810T150000 - 20190810T160000 FOO BAR BAZ [@2]
20190810T160000 - 20190810T170000 FOO         [@1]
```
the resulting behaviour would be
* `continue FOO` continues interval `@1` 
* `continue FOO BAR` would continue interval `@2` (although `@3` would be an exact match)
* `continue BAR` would also continue `@2`

Which also means there would be no way to continue `@4` by tags

When you are already specifying an id, then I would argue there is no need to specify any tags (ids trump tags). The id is *the* reference to the to be continued interval. So 
```
timew continue @3 FOO BAR
```
could either ignore the supplied tags or return an error ("you cannot specify both id and tags to continue an interval"). I opt for the latter to avoid unexpeced behaviour.

lauft:
> I'm not quite sure if you meant it like that regarding the filter...

I meant the filtering by tag is already integrated in `getTracked` (src/timew.h:54).
`getFilter(const CLI&)` should return the tags from the command line and `getTracked` filter out the intervals not matching them (see src/data.cpp:540)

croesch:
Thanks for the comments, regarding your behaviour definition, I agree with it. Currently no error occurs, but I can change that.

> I meant the filtering by tag is already integrated in `getTracked` (src/timew.h:54).
> `getFilter(const CLI&)` should return the tags from the command line and `getTracked` filter out the intervals not matching them (see src/data.cpp:540)

Yes, I'm still not 100% sure - could you review the refactoring if it matches your expectations? If yes, I'll integrate the error for tags and id at one time.

xerus2000:
Please get back to this, I'd love this feature!

lauft:
@croesch I have finished your PR and commited it in f3280966e0f21a6a219f71f361bd1e290e85ef79..d99872eb9084e3510971c1f3c8053afbf3c7a1f8.
Thanks again for your contribution!

PR #261, opened 2019-11-13
Double quotes in annotations and tags not handled.

sruffell:
A potential fix for handling quotation marks in annotations.


lauft:
Merged. Thanks for your contribution

PR #262, opened 2019-11-18
Allow tests to use manpages in source tree

sruffell:
Currently, since `test/help.t` uses `system("man timew-shorten")` to check, the man pages need to be installed in the system.

This change makes it so that the test can use a custom MANPATH, so that the local man pages are used, and not the ones that are installed in the system.


test/help.t
```diff
@@ -61,7 +61,10 @@ class TestHelp(TestCase):
     def test_help_with_command_should_show_man_page(self):
         """timew help with command should show man page"""
         code, out, err = self.t("help start")
-        self.assertEquals(err, "")
+        # Some versions of man have bug that displays following on stderr:
+        #         doc/man1/timew-shorten.1.in: ignoring bogus filename
+        # We will just check that stdout is correct.
+        #self.assertEquals(err, "")
```
lauft: If this line is not needed anymore, better delete it.

PR #270, opened 2019-12-23
Stop with id should suggest modify command

sruffell:
Recently I was trying to stop an interval that I had left open over the
weekend, but I had already started a new interval at the beginning of the
week. The error I was received was "The end of a date range must be after the
start." which was confusing to me.

After a few moments I realized I should have been using the modify command.
Now the stop command will suggest this if someone tries.

### DIFF
--- a/src/commands/CmdStop.cpp
+++ b/src/commands/CmdStop.cpp
@@ -50,10 +50,16 @@ int CmdStop (
   // Load the most recent interval.
   auto filter = getFilter (cli);
   auto latest = getLatestInterval (database);
+  std::set <int> ids = cli.getIds ();
 
   // Verify the interval is open.
   if (! latest.is_open ())
     throw std::string ("There is no active time tracking.");
+
+  // We either expect no ids, or we're operating on the most current.
+  if (! ids.empty ())
+    throw std::string ("The stop command works on the most recent open interval. "
+                       "Perhaps you want the modify command?.");
 
   journal.startTransaction ();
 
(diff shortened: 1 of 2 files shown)


test/stop.t
```diff
@@ -198,6 +198,18 @@ class TestStop(TestCase):
...
+    def test_stop_with_id(self):
+        """Stop does not work with with ids other than one"""
+        self.t("start 2h ago")
+        self.t("start 1h ago")
+        
+        self.t("stop @1")
+
+        code, out, err = self.t.runError("stop @2")
+        # If trying to stop an older interval, check that modify was
+        # suggested.
```
lauft: There is no 80-char-linelength-rule, use the space. If, then it is 'one sentence, one line' :)

src/commands/CmdStop.cpp
```diff
@@ -50,6 +50,15 @@ int CmdStop (
...
   auto latest = getLatestInterval (database);
+  std::set <int> ids = cli.getIds ();
+
+  // We either expect no ids, or we're operating on the most current.
+  for (auto id : ids)
+  {
+    if (id != 1)
+      throw std::string ("Stop only works on the most recent open interval. "
+                         "Perhaps you want the modify command?.");
+  }
```
lauft: I would simplify the check to a single `ids.empty ()`.

sruffell: Done in sruffell/timewarrior@c9325f1.

test/stop.t
sruffell: Done in sruffell/timewarrior@c9325f1

Issue #259, opened 2019-11-12
Migrate to python 3


lauft:
With python 2 reaching its end of life (see https://pythonclock.org/), it is time to move on...
Most/all python files should run with python 3 already but now it should be made explicit.

lauft:
ToDos:
* Migrate tests to python 3 (-> adapt setup for docker containers!)
* Migrate hook scripts to python 3 (-> extract to separate repository?)

lauft:
Current status:
Done:
* test suite migrated to python3
* Travis runs all tests on python3
To do:
* Migrate hook script `on-modify.timewarrior` to python 3
* Migrate reports to python 3

sruffell:
@lauft Just a note that `test/problems` has not yet been switched.

PR #274, opened 2020-01-04
259: Upgrade on-modify.timewarrior to python3

tom-doerr:
Regarding #259.
The hook script seems to be compatible to python3 without further modifications.


lauft:
Done in #259

PR #275, opened 2020-01-08
Change `timew help intervals` to `timew help ranges` in help summary

sruffell:
These are minor changes to fix the strings displayed to the user for obtaining additional help.

### DIFF
--- a/test/help.t
+++ b/test/help.t
@@ -68,7 +68,7 @@ class TestHelp(TestCase):
 
     def test_help_with_unknown_argument_should_show_error_message(self):
         """timew help with unknown argument should show error message"""
-        code, out, err = self.t("help bogus")
+        code, out, err = self.t.runError("help bogus")
         self.assertRegex(err, r"No manual entry for timew-bogus")
 
     def test_command_with_help_long_option_should_show_help_page(self):
--- a/src/commands/.gitignore
+++ b/src/commands/.gitignore
@@ -1,1 +1,2 @@
 libcommands.a
+additional-help.h
--- a/src/commands/CMakeLists.txt
+++ b/src/commands/CMakeLists.txt
@@ -5,6 +5,18 @@ include_directories (${CMAKE_SOURCE_DIR}
                      ${CMAKE_SOURCE_DIR}/src/libshared/src
                      ${TIMEW_INCLUDE_DIRS})
 
+set (ADDITIONAL_HELP_H "additional-help.h")
+
+ADD_CUSTOM_COMMAND (
+	OUTPUT ${ADDITIONAL_HELP_H}
+	COMMAND /bin/sh src/commands/generate-additional-help.sh >src/commands/${ADDITIONAL_HELP_H}
+	DEPENDS ${CMAKE_CURRENT_LIST_DIR}/generate-additional-help.sh
+	WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+	COMMENT "Generating ${ADDITIONAL_HELP_H}")
+
+ADD_CUSTOM_TARGET (generate_additional_help
+	DEPENDS ${ADDITIONAL_HELP_H})
+
 set (commands_SRCS CmdAnnotate.cpp
                    CmdCancel.cpp
                    CmdChart.cpp
@@ -38,4 +50,4 @@ set (commands_SRCS CmdAnnotate.cpp
                    CmdUntag.cpp)
 
 add_library (commands STATIC ${commands_SRCS})
-
+add_dependencies (commands generate_additional_help)
--- a/src/commands/CmdHelp.cpp
+++ b/src/commands/CmdHelp.cpp
@@ -117,8 +117,8 @@ int CmdHelp (
   if (! words.empty ())
   {
     std::string man_command = "man timew-" + words[0];
-    system (man_command.c_str());
-    return 0;
+    int ret = system (man_command.c_str());
+    return (WIFEXITED (ret)) ? WEXITSTATUS (ret) : -1;
   }
 
   return CmdHelpUsage (extensions);
--- a/src/commands/CmdHelp.cpp
+++ b/src/commands/CmdHelp.cpp
@@ -26,13 +26,67 @@
 
 #include <cmake.h>
 #include <commands.h>
+#include <algorithm>
+#include <cstring>
 #include <iostream>
 #include <stdio.h>
 #include <FS.h>
+#include <sstream>
+
+#include "additional-help.h"
+
+template <typename T>
+std::string join (const T& strings, size_t indent = 0, size_t max_width = 0)
+{
+  std::stringstream sstr;
+  auto it = std::begin (strings);
+  auto end = std::end (strings);
+  size_t cur = indent + 2; // account for added '{ ' 
+  std::string part;
+  const std::string sep (" | ");
+
+  while (it != end)
+  {
+    part = *it;
+    ++it;
+
+    // Make sure we leave room for closing " }" or " |"
+    if ((max_width > 0) && (cur + part.length () + sep.length ()) > max_width)
+    {
+      sstr << "\n" << std::string (indent + 1, ' ');
+      cur = indent + 1;
+      sstr << part;
+    }
+    else
+    {
+      sstr << part;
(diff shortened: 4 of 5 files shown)


test/help.t
```diff
@@ -83,6 +84,21 @@ class TestHelp(TestCase):
         code, out2, err2 = self.t("track -h")
         self.assertEqual(out1, out2)
 
+    def test_help_additional_info(self):
+        """Test that there are man pages for all the additional help pages."""
```
lauft: I would try to keep the test naming scheme "timew ... should ...", e.g. "timew should display additional help pages"

test/help.t
```diff
@@ -83,6 +84,21 @@ class TestHelp(TestCase):
...
+
+        additional_help = re.compile(r".*timew (help [a-x]+)\s*")
+
+        additional_help_count = 0
+        code, out, err = self.t("help")
```
lauft: Nice idea to use timew to get the current additional help items, but why not simply scan the `doc/man[0-9]` directory for `timew-(.+)\.in` and use this match to test the help command.

This way we would test all current possible help arguments, *including* the commands. :wink:

With this extension this test would become "timew should display help pages" - and would replace the narrower `test_help_with_command_should_show_man_page` from `help.t:62`

sruffell: @lauft, thanks for the review.

> Nice idea to use timew to get the current additional help items, but why not simply scan the doc/man[0-9] directory for timew-(.+)\.in and use this match to test the help command.

I can see how this would be useful to make sure all the manpages in the source are compiled into something that can be installed, but I'm not sure scanning the directory will achieve the goal of ensuring that these strings embedded in commands/CmdHelp.cpp actually map to man pages.

Below are the strings on 1.2.0 branch, and you can see how it suggests that `timew help interval` be used for additional information, but 
https://github.com/GothenburgBitFactory/timewarrior/blob/1d271508a27b9d8a5211ae5b01c7a0583ff690fd/src/commands/CmdHelp.cpp#L79-L85

But in 4c19a055f4f7a751, the file was renamed so that `timew help interval` does not work anymore.  It was the output of `timew help` that I was specifically trying to create a test for.

Am I missing your point?

test/help.t
sruffell: I should have noticed I wasn't following the pattern. I'll update and resolve this conversation when I update this pull request pending the question about what to actually test for below.

test/help.t
lauft: I think there are two aspects to consider here:
* does `timew help <argument>` work for every valid `<argument>`?
* does `timew help` produce valid output?
maybe a variant of the second:
* does `timew help` contain all valid options?

Question 1 is answered quite easily by the method I proposed - if you define "valid `<argument>`" as "everything there is a man page for".

The error you detected touches more question 2 (and 3): The output of `timew help` did not contain the new name of the man page.

We could have test for both scenarios, but I consider tests that parse output as brittle an thus try to avoid them. I also think a full test for the validity of the output `timew help` is unfeasible*... we cannot test away proof-reading the documentation.

* it may be possible if the output is generated (and not hardcoded like now) - then we can test the generator.

sruffell: > * it may be possible if the output is generated (and not hardcoded like now) - then we can test the generator.

I've taken your suggestion and updated the pull request to generate the commands for the Additional help section from the *.in files in the docs directory.

It now looks like:
```
Additional help:
       timew help [COMMAND | CONCEPT]

       COMMAND = { annotate | cancel | chart | config | continue | day |
                   delete | diagnostics | export | extensions | fill | gaps |
                   get | help | join | lengthen | modify | month | move |
                   report | resize | shorten | show | split | start | stop |
                   summary | tag | tags | track | undo | untag | week }

       CONCEPT = { config | dates | dom | durations | hints | ranges }
```

Is something like this acceptable?

lauft: With this we would parse the man directories, generate the help page from it, then parse the generated help page to determine for which commands/concepts to test `timew help` for. 

This is one level too much for the test, where we can use the directory parsing output directly -- which would then also avoid the need to adapt the parsing whenever the display format is changed. The tests could look like this: 
```python
def test_help_should_show_documented_commands(self):
    """timew help should show documented commands"""
    regex = re.compile(r"timew-(.+)\.1\.in", re.MULTILINE)

    for root, dirs, files in os.walk("../doc/man1/."):
        for file in files:
            match = regex.search(file)
            if match:
                command = match.group(1)
                code, out, err = self.t("help {}".format(command))
                self.assertRegex(out, r"timew-.+\(1\)\s+User Manuals\s+timew-.+\(1\)".format(command))

def test_help_should_show_documented_concepts(self):
    """timew help should show documented concepts"""
    regex = re.compile(r"timew-(.+)\.7\.in", re.MULTILINE)

    for root, dirs, files in os.walk("../doc/man7/."):
        for file in files:
            match = regex.search(file)
            if match:
                concept = match.group(1)

                if concept == "config":
                    # Not accessible via `timew help config` which shows help for command `config`
                    continue

                code, out, err = self.t("help {}".format(concept))
                self.assertRegex(out, r"timew-.+\(7\)\s+User Manuals\s+timew-.+\(7\)".format(concept))
``` 
However we need to exclude `config` from the tested concepts as `timew help config` will show the man page for the command with the same name.

I would also keep the format of the additional help section as it was before and only generate the lines for the documented concepts, changing your code to something like this:
```c++
...
std::cout << "Additional help:\n"
          << "       timew help <command>\n";
for (auto concept: documented_concepts)
{
  if (concept == "config")
  {
    // Not accessible via `timew help config` which shows help for command `config`
    continue;
  }
  std::cout << "       timew help " << concept << '\n';
}
...
```
The downside here is (again) the exception for `config`... 🤔

sruffell:
TLDR; You are OK with generating the output of the help page for the concepts (and filtering config), but you would like the test to just check the man pages directly and not parse the output? If so, I can update this with that change.

Longer form below... 

> With this we would parse the man directories, generate the help page from it, then parse the generated help page to determine for which commands/concepts to test timew help for.
>
> This is one level too much for the test, where we can use the directory parsing output directly -- which would then also avoid the need to adapt the parsing whenever the display format is changed. 

Ok. My thought on this was that it was better to require a developer to update the parsing code than potentially shipping a release with a command that didn't work. I figured if the format was changed to something that was easily parsed and primarily generated the risk of an inadvertent change would be low.  The format I adopted was like those from the `ip` command.

```shell
NAME
       ip - show / manipulate routing, network devices, interfaces and tunnels

SYNOPSIS
       ip [ OPTIONS ] OBJECT { COMMAND | help }

       ip [ -force ] -batch filename

       OBJECT := { link | address | addrlabel | route | rule | neigh | ntable | tunnel | tuntap | maddress | mroute | mrule | monitor | xfrm | netns | l2tp | tcp_metrics | token | mac‐
               sec }
...
```

lauft:
As a compromise one could go twofold:
* testing whether `timew help { COMMAND | CONCEPT }` works by parsing the dir content, calling the respective commands
* testing whether `timew help` produces "valid" output by pasing its output and run some validation on it.

However I consider the second test brittle. It depends on *external* third party tools (different operating systems, man pager, ...) and on how we format the page itself. The result is - depending on the parsing - just a "yes, there is a FOO somewhere in there" to "line 123 contains exacly '...FOO...'". But the more precise, the more it becomes susceptible to (outside) influences. It may work, but it may also become a maintenance burden. 🤔 

In any case it does not free us from proofreading the man pages (and other documentation). Just because we know "FOO is in there" does not mean the documentation is correct. 🤷‍♂ 

So... I guess *if* such a test is to be added, it should be as simple and minimal as possible, but then it will only serve as a beacon for possible errors.

Another thought: I think it is a bad idea to change the documentation to "...something that was easily parsed..." just to be able to test it better. Documentation should be _readable_ not _parseable_.

sruffell:
@lauft Since I see we're getting close to having completed all the 1.2.1 milestone I thought I would go ahead and resubmit this, but without the test that parses the output of `timew help`. 

All this currently does is generate a vector of all the concepts (that aren't also a command) and use that vector in the output of `timew help`.

One other question I had though, is that <interval> is used on the help page liberally, but should the majority of those be changed to <range> to match the concept?

PR #286, opened 2020-02-11
285: Pass interval id to extensions

### LINKED ISSUE #285: Pass record id to extensions
Currently, extensions only receive start, end, tags and annotations through stdin. I am currently building a extension that sends certain entries to the tempo jira plugin (for work). As such I need to somehow track which entries have already been sent to tempo and which ones have not.

From my tests I have not yet found a way to get from a JSON line in an extension to the record in order to update it.

I am not 100% comfortable with C++ - but I will try to have a stab at implementing this since it sounds fairly straight forward.

j6s:
Passes the `id` of an interval to extensions. Now, stdin of an extension looks like this:

```
{"id":99,"start":"20191219T183718Z","end":"20191219T185724Z","annotation":"project:gaming"},
{"id":98,"start":"20191219T211035Z","end":"20191219T220034Z","tags":["project:gaming"]},
{"id":97,"start":"20200107T090000Z","end":"20200107T113235Z","tags":["tempo:ABC-477"]},
{"id":96,"start":"20200107T125607Z","end":"20200107T152439Z","tags":["tempo:ABC-477"]},
{"id":95,"start":"20200107T165000Z","end":"20200107T180803Z","tags":["tempo:ABC-477"]},
{"id":94,"start":"20200108T081022Z","end":"20200108T083142Z","tags":["tempo:ABC-477"]},
```

Resolves #285

### DIFF
--- a/src/Interval.cpp
+++ b/src/Interval.cpp
@@ -99,16 +99,14 @@ std::string Interval::serialize () const
 std::string Interval::json () const
 {
   std::stringstream out;
-  out << '{';
+  out << "{\"id\":" << id;
 
-  if (is_started ())
-    out << "\"start\":\"" << start.toISO () << "\"";
+  if (is_started ()) {
+    out << ",\"start\":\"" << start.toISO () << "\"";
+  }
 
-  if (is_ended ())
-  {
-    if (is_started ())
-      out << ',';
-    out << "\"end\":\"" << end.toISO () << "\"";
+  if (is_ended ()) {
+    out << ",\"end\":\"" << end.toISO () << "\"";
   }
 
   if (! _tags.empty ())


src/Interval.cpp
```diff
@@ -99,17 +99,13 @@ std::string Interval::serialize () const
 std::string Interval::json () const
 {
   std::stringstream out;
-  out << '{';
+  out << "{\"id\":" << id;
 
   if (is_started ())
-    out << "\"start\":\"" << start.toISO () << "\"";
+    out << ",\"start\":\"" << start.toISO () << "\"";
```
sruffell: Just a note that going forward @lauft has asked for there to be curly braces around all code blocks in new and modified code.

See:
https://github.com/GothenburgBitFactory/timewarrior/pull/269#discussion_r367937920

src/Interval.cpp
```diff
@@ -99,17 +99,13 @@ std::string Interval::serialize () const
...
 
   if (is_ended ())
-  {
-    if (is_started ())
-      out << ',';
-    out << "\"end\":\"" << end.toISO () << "\"";
-  }
+    out << ",\"end\":\"" << end.toISO () << "\"";
```
sruffell: Here too you'll want to add curly braces.

src/Interval.cpp
j6s: Wasn't aware of that, will change.

I saw the one statement without braces and thought that was the way this project did them.

src/Interval.cpp
lauft: I know that there rarely is an interval without a start, but I would prefer to keep lines 109 - 110 for consitency

j6s: As I understand lines 109-110 were adding a comma in front only if the start existed because if a start existed, the start would be the first item in the json object but if a start did not exist, the end time would be the first item in the object.

With this change, there is always a defined first item in the object: the id.

lauft: @j6s you are definitely right here 😅 - however I found that there is the case of an empty interval which then gets serialized as `{"id":0}` instead of `{}`.
And also tests `dom.t` and `interval.t` are broken. They appear green on Travis (should be fixed with https://github.com/GothenburgBitFactory/timewarrior/commit/dcfe0d1c), but if you inspect the individual runs you see the error reports (e.g. https://travis-ci.org/GothenburgBitFactory/timewarrior/jobs/651153570)

lauft:
I discovered that this feature comes handy for tests, also in context of #294. I will merge it and handle the open issues mentioned in the review. Thanks @j6s nevertheless for this contribution!

j6s:
Sorry for not finishing this - I had genuinely planned to have a look at it on the weekend.

lauft:
@j6s Don't be. You can take a look at the documentation and see whether (or better where) this needs to be updated. And, also important, consider to publish your addon on github/... and tell us about it so we can mention it to other users 😉

PR #299, opened 2020-03-08
Do not display misleading output on errors

### LINKED ISSUE #280: Should not output "Recorded" message when stopping fails
Just did the following typo (should have been "15min" instead of  "15 min"):

```
> timew stop 15 min ago                                                                                                                                    
Recorded XXX
  Started 2020-01-30TXX:XX:XX
  Ended              XX:XX:XX
  Total               XX:XX:XX
The current interval does not have the '15' tag.
```

The error message is correct, and the interval was not stopped, which is also correct. 

But the bug here is that there is a "Recorded" message printed, which indicates that tracking XXX has been stopped. That's maybe related to #116.

sruffell:
This eliminates possibly misleading messages from the console. For example, in
the following before and after, the after eliminates the message that 'test' was
recorded when it was not due to the user error:

Before this change:

  $ timew stop 15 min ago
  Recorded test
    Started 2020-03-08T13:11:20
    Ended                 40:31
    Total               0:29:11
  The current interval does not have the '15' tag.

After this change:

  $ timew stop 15 min ago
  The current interval does not have the '15' tag.

Closes #280

### DIFF
--- a/src/Database.cpp
+++ b/src/Database.cpp
@@ -307,7 +307,7 @@ void Database::addInterval (const Interval& interval, bool verbose)
   {
     if (_tagInfoDatabase.incrementTag (tag) == -1 && verbose)
     {
-      std::cout << "Note: '" << quoteIfNeeded (tag) << "' is a new tag." << std::endl;
+      timew::cout << "Note: '" << quoteIfNeeded (tag) << "' is a new tag." << std::endl;
     }
   }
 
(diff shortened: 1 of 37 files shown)


sruffell:
There aren't any significant changes in the update. I just rebased so it would apply cleanly to the dev branch.

sruffell:
Again, I rebased and fixed up conflicts.

@lauft let me know if this is something you're interested in or have another idea about how to resolve #280.

lauft:
@sruffell Sorry, this PR got somehow buried underneath other stuff...

When taking a second look at the problem, I found it simpler to just move all the validation (filter range, tag set, ...) up before any database actions are taken (see 5b1c77b8aa3d9f723b45aae62cf168f7ae750a0b, 96aaa71eb40c02c3e28ceb5ad2f50592eba24686, and 3feba61459b2e21c8caa55257dae731d9050d3fa).

PR #312, opened 2020-05-05
Follow XDG

### LINKED ISSUE #207: Implement XDG Base Directory Specification
Adhering to the XDG Base Directory Specification may store Timewarrior data and configuration in a more standardized way. This may resolve some topics mentioned in #193.

Implementing the specification would mean to
* move configuration to  `$XDG_CONFIG_HOME/timewarrior` (defaults to `$HOME/.config/timewarrior`)
* move data to `$XDG_DATA_HOME/timewarrior` (defaults to `$HOME/.local/share/timewarrior`)

Coming with this are some questions:
* What about users who do multiple time tracking and separate this via `$TIMEWARRIORDB`? Is this a relevant use case?
* How to migrate data from `$HOME/timewarriordb`? Separate script, built-in?
* Is the benefit worth the effort?

xeruf:
Still a WIP, right now it merely hardcodes the defaults.

Fixes #207

### DIFF
--- a/src/Database.h
+++ b/src/Database.h
@@ -120,7 +120,7 @@ private:
   void initializeTagDatabase ();
 
 private:
-  std::string               _location {"~/.timewarrior/data"};
+  std::string               _location {"~/.local/share/timewarrior/data"};
   std::vector <Datafile>    _files    {};
   TagInfoDatabase           _tagInfoDatabase {};
   Journal*                  _journal {};
--- a/src/Journal.h
+++ b/src/Journal.h
@@ -52,7 +52,7 @@ public:
 private:
   void recordUndoAction (const std::string &, const std::string &, const std::string &);
 
-  std::string _location {"~/.timewarrior/data/undo.data"};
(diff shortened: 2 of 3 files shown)


lauft:
@xerus2000 I may state the obvious, but hardcoding is not an option. We cannot just switch from one path scheme to another, we need to migrate the existing users - if they want to. 

Therefore it is essential that we supply a way to keep the old setup after the change (e.g. by overriding the defauts for the `XDG_...` environment variables) or a way to transition smoothly to the new structure...

xerus2000:
I am aware of all that, I just had this draft sitting around and wanted to make sure it doesn't get lost ;)

sruffell:
@xerus2000 are you planning on updating this? This seems like something timewarrior should have for the 2.0.0 release.

xerus2000:
@sruffell I'd love to, but I am pretty inexperienced in C, and currently low on time, so it will likely take a while.

lauft:
Done in #478

PR #330, opened 2020-05-26
#206 add all hint

### LINKED ISSUE #206: Add :all hint
It would be great to have a `:all` hint which could be used to display all intervals over the years (or is there already a function for this that I'm not aware of?).

sruffell:
This adds the :all hint and also updates CmdSummary and CmdGaps to run more efficiently given the large interval.

Closes #206

### DIFF
--- a/src/timew.h
+++ b/src/timew.h
@@ -40,9 +40,6 @@
 std::vector <Range>     getHolidays       (const Rules&);
 std::vector <Range>     getAllExclusions  (const Rules&, const Range&);
 std::vector <Interval>  getIntervalsByIds (Database&, const Rules&, const std::set <int>&);
-std::vector <Interval>  subset            (const Interval&, const std::vector <Interval>&);
-std::vector <Range>     subset            (const Range&, const std::vector <Range>&);
-std::vector <Interval>  subset            (const Range&, const std::vector <Interval>&);
 void                    flattenDatabase   (Database&, const Rules&);
 std::vector <Interval>  flatten           (const Interval&, const std::vector <Range>&);
 std::vector <Range>     merge             (const std::vector <Range>&);
--- a/src/timew.h
+++ b/src/timew.h
@@ -98,4 +98,26 @@ std::string joinQuotedIfNeeded(const std::string& glue, const std::vector <std::
 // dom.cpp
 bool domGet (Database&, Interval&, const Rules&, const std::string&, std::string&);
 
+////////////////////////////////////////////////////////////////////////////////
+// Given a sorted collection of ranges, return a pair of iterators for all
+// elements that intersect the supplied day range. The end iterator will be the
+// first interval that does not intersect the range or the passed end
+//
+template <typename It>
+std::pair <It, It> intersects (It begin, It end, const Range& day_range)
+{
+  while (begin != end && ! day_range.intersects (*begin) && begin->start < day_range.start)
+  {
+    ++begin;
+  }
+
+  It iend = begin;
+  while (iend != end && day_range.intersects (*iend))
+  {
+    ++iend;
+  }
+
+  return std::pair<It, It> {begin, iend};
+}
+
 #endif
--- a/test/summary.t
+++ b/test/summary.t
@@ -223,6 +223,28 @@ W\d{1,2} \d{4}-\d{2}-\d{2} .{3}       ?0:00:00 0:00:00 0:00:00 0:00:00
 [ ]+0:00:00
 """)
 
+    def test_with_multiple_multi_day_tracks(self):
+        """Summary should display intervals that span multiple days"""
+        self.t("track sopy for 3d proja")
+        self.t("track sopm for 3d projb")
+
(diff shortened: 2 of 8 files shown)


src/timew.h
```diff
@@ -57,6 +54,11 @@ std::vector <Range>     getUntracked      (Database&, const Rules&, Interval&);
 Interval                getLatestInterval (Database&);
 Range                   getFullDay        (const Datetime&);
 
+template <typename It>
+std::pair <It, It>      intersects        (It begin, It end, const Range& day_range);
+
+#include <data.inl>
```
lauft, line 60: Why not include the template code here? Just like in `libshared/src/format.h` for example

src/helper.cpp
```diff
@@ -151,6 +151,8 @@ bool expandIntervalHint (
     {":month",       {"som",       "eom"}},
     {":quarter",     {"soq",       "eoq"}},
     {":year",        {"soy",       "eoy"}},
+    // Use the "epoch" from the Datetime class, which is 1980-01-01
+    {":all",         {"315532800", "eod"}},
```
lauft, line 155: Time tracking before 1980 and/or in the future is unlikely (and forbidden in the latter case) but I think it is better to expand the `:all` hint  into what it really means, i.e. an open range, instead of introducing artificial boundarys.

test/summary.t
```diff
@@ -223,6 +223,28 @@ W\d{1,2} \d{4}-\d{2}-\d{2} .{3}       ?0:00:00 0:00:00 0:00:00 0:00:00
...
+        # W1 1980-01-01 Tue @2 proja 1:00:00 0:00:00 23:00:00 23:00:00
+        # W1 1980-01-02 Wed @2 proja 0:00:00 0:00:00 24:00:00 24:00:00
+        # W1 1980-01-03 Thu @2 proja 0:00:00 1:00:00  1:00:00  1:00:00
+        # W1 1990-01-01 Mon @1 projb 1:00:00 0:00:00 23:00:00 23:00:00
+        # W1 1990-01-02 Tue @1 projb 0:00:00 0:00:00 24:00:00 24:00:00
+        # W1 1990-01-03 Wed @1 projb 0:00:00 1:00:00  1:00:00  1:00:00
+        #
+        #                                                     96:00:00
+
+        code, out, err = self.t("summary :ids 1980-01-01T00:00:00 - eoy")
```
lauft, line 245: Why not use the `:all` hint here? 😉

test/summary.t
```diff
@@ -223,6 +223,28 @@ W\d{1,2} \d{4}-\d{2}-\d{2} .{3}       ?0:00:00 0:00:00 0:00:00 0:00:00
 [ ]+0:00:00
 """)
 
+    def test_with_multiple_multi_day_tracks(self):
+        """Summary should display intervals that span multiple days"""
+        self.t("track 1980-01-01T01:00:00 - 1980-01-03T01:00:00 proja")
```
lauft, line 228: I would prefer to use relative dates instead of fixed ones, so one has an extra bit of randomness in the test data. 
Fixed dates are easier to debug but you do not have the chance to spot those errors that only occur when your test date is X days after special date Y...

sruffell:
> But I think the `:all` hint should expand into an emtpy range (`Range {0, 0}`), i.e. an infinite open interval, just like when all data is requested (e.g. `src/commands/CmdFill.cpp:49`).

One issue with using an open range is how CmdSummary will auto close open intervals to span a single day.  It isn't clear to me how you would like CmdSummary to handle the differences between `timew summary yesterday` and `timew summary :all` in terms of knowing when auto closing an open interval at a single day is preferred versus leaving it open.

test/summary.t
sruffell: Eh, I wrote that test against master when I was curious if I added a bug as part of adding the all hint. I will adjust it in the eventual updated pull request.

src/helper.cpp
sruffell: Defining it as an open range does create a problem with CmdSummary. But perhaps the filter needs to auto close a single day to be closed interval itself and not leave that to CmdSummary?

src/timew.h
sruffell: I thought that your preference wasn't to add implementations in the .h files, but I can move it when this is updated.

src/helper.cpp
sruffell: I just modified CmdSummary to only auto close the interval for a day if it was already started.

lauft:
>  It isn't clear to me how you would like CmdSummary to handle the differences between timew summary yesterday and timew summary :all in terms of knowing when auto closing an open interval at a single day is preferred versus leaving it open.

I think `CmdSummary` (and all the other commands) should not handle those differences at all!
`CmdSummary` should request the date range from the CLI and query and display the intervals accordingly. (This means the handling has to be moved to `CLI::getFilter`.)

Also see #333 for this

sruffell:
Sounds good. I'll close this pull request for the time being and can create a new one for any performance/test improvements after seeing how that shakes out.

PR #331, opened 2020-05-26
#326 start command adjust

### LINKED ISSUE #326: start command with :adjust does not overwrite open interval
I already had a task running,but I realised it was the wrong one. It was started 10:50:05, so I typed

    ❯ timew start 10:50 mixxx
    The end of a date range must be after the start.

That error is already rather unhelpful, as it doesn't mention that the clash is due to the previous task.
But the more annoying problem is that even when specifying the adjust hint, I got that same error. I would expect to simply override the previous interval if that hint is supplied.

sruffell:
Closes #326

### DIFF
--- a/src/commands/CmdStart.cpp
+++ b/src/commands/CmdStart.cpp
@@ -48,12 +48,17 @@ int CmdStart (
 
   journal.startTransaction ();
 
+  if (!filter.is_started ())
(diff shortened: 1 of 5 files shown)


sruffell:
@lauft has taken care of this already in #326.

PR #335, opened 2020-06-11
Update docs links from taskwarrior.org to timewarrior.net

georgeblahblah:
Signed-off-by: George Buckingham <george.buckingham@gmail.com>


lauft:
Thanks.

PR #363, opened 2020-08-13
CLI: Reserve arguments that start with ':' for hints

### LINKED ISSUE #348: Should arguments that begin with ':' be reserved for hints?
The following output surprised me:

```
$ src/timew track :yesterday foo
Note: 'foo' is a new tag.
Recorded foo
  Started 2020-07-23T00:00:00
  Ended           24T00:00:00
  Total              24:00:00
$ src/timew track :today bar
Note: ':today' is a new tag.
Note: 'bar' is a new tag.
Tracking :today bar
  Started 2020-07-24T14:43:52
  Current                  52
  Total               0:00:00
```

My question is should any command line argument that begins with a `:` be interpreted as a hint, and display an error that there is no hint with that name instead of treating it as a tag?

sruffell:
Currently, any argument that starts with ':' that isn't a known hint
will be treated as a tag. After this change, an error is produced.

Hopefully there were not users who were using tags that start with a
colon...

Closes #348


test/cli.t
```diff
@@ -103,6 +103,11 @@ class TestCLI(TestCase):
         code, out, err = self.t.runError("bogus")
         self.assertIn("'bogus' is not a timew command. See 'timew help'.", err)
 
+    def test_TimeWarrior_with_invalid_hint(self):
+        """Call a non-existing TimeWarrior hint should be an error"""
+        code, out, err = self.t.runError("start :invalid_hint")
```
lauft, line 108: wouldn't `timew :unknown` (without the `start` command) also work? 🤔

lauft (commented):
Lookin good so far, but I think we need also to address cases when a user tries to `(un)tag` an interval with a (pseudo-)hint, e.g.:
```
timew (un)tag @1 :quiet # should already be handled...
timew (un)tag @1 :unknown
timew (un)tag @1 FOO :unknown 
```

test/cli.t
sruffell: Removed start in latest push.

src/CLI.cpp
```diff
@@ -465,11 +465,18 @@ void CLI::canonicalizeNames ()
...
-      a.tag ("HINT");
+      if (exactMatch ("hint", raw) ||
+          canonicalize (canonical, "hint", raw))
+      {
+        a.attribute ("canonical", canonical);
+        a.tag ("HINT");
+      }
+      else
+      {
+        throw format ("'{1}' is an invalid hint.", raw);
```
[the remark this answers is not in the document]
sruffell: Updated in latest push.

sruffell:
> Lookin good so far, but I think we need also to address cases when a user tries to `(un)tag` an interval with a (pseudo-)hint, e.g.:
> 
> ```
> timew (un)tag @1 :quiet # should already be handled...
> timew (un)tag @1 :unknown
> timew (un)tag @1 FOO :unknown 
> ```

I'm not sure I understand?  Anytime an invalid hint is specified, timewarrior will report the error halt.   What would you like to happen.

Here is the behavior in the current iteration of this pull request:
```
$ src/timew tag @1 :unknown
Unknown hint ':unknown'
$ src/timew tag @1 FOO :unknown
Unknown hint ':unknown'
```

Issue #308, opened 2020-04-28
`timew` command reports incorrect running time in cygwin


daytonb:
When I run the `timew` command, I often get the wrong amount of running time on my Windows computer where I compiled and use timew in a cygwin terminal. To confirm the issue, I compiled on a CentOS 8 machine using the same build commands. Also, the time is correctly reported when I run `timew sum`.

Windows machine (cygwin)
```
$ ~/opt/timew-test-install/bin/timew.exe
Tracking sapphire
  Started 2020-04-28T06:25:00
  Current               58:42
  Total               2:31:41

$ ~/opt/timew-test-install/bin/timew.exe sum

Wk  Date       Day Tags       Start End    Time   Total
W18 2020-04-28 Tue sapphire 6:25:00   - 0:33:44 0:33:44

                                                0:33:44


$ ~/opt/timew-1.1.1/bin/timew.exe diagnostics

timew 1.1.1
       Platform: Cygwin

Compiler:
        Version: 7.4.0
           Caps: +stdc +stdc_hosted +200809 +200809 +LP64 +c8 +i32 +l64 +vp64 +time_t64
     Compliance: C++11

Build Features
          Built: Jan 29 2020 10:59:44
          CMake: 3.14.5
     Build type: release

Configuration
  TIMEWARRIORDB: -
            Cfg: /home/username/.timewarrior/timewarrior.cfg (-rw- 231 bytes)
       Database: /home/username/.timewarrior (drwx 0 bytes)
    Color theme: Built-in default
                 00 01 02 03 04 05 06 07 08 09 10 11 12

Extensions
       Location: /home/username/.timewarrior/extensions (drwx 0 bytes)
               (None)
```

Linux machine
```
$ ~/opt/timew-1.1.1/bin/timew
Tracking
  Started 2020-04-28T06:25:00
  Current               59:43
  Total               0:34:43
$ ~/opt/timew-1.1.1/bin/timew sum

Wk  Date       Day Tags   Start End    Time   Total
W18 2020-04-28 Tue      6:25:00   - 0:34:49 0:34:49

                                            0:34:49

$ ~/opt/timew-1.1.1/bin/timew diagnostics

timew 1.1.1
       Platform: Linux

Compiler:
        Version: 8.3.1 20190507 (Red Hat 8.3.1-4)
           Caps: +stdc +stdc_hosted +LP64 +c8 +i32 +l64 +vp64 +time_t64
     Compliance: C++11

Build Features
[...]

PR #368, opened 2020-08-14
Use only the current interval duration when summarizing

### LINKED ISSUE #248: Inconsistent current activity duration
Often, timewarrior gives an incorrect duration of the current activity if the activity is first of the given day.

I believe, this issue is best described by a snapshot of the output.

    $ timew
    Tracking xxxx
      Started 2019-08-27T10:05:05
      Current               35:38
      Total               0:57:14

sruffell:
There are several instances where the "total" reported in the interval
summary is confusing because the summarize function would walk backward
and add all intervals with the same set of tags.

Closes #248 and #308

### DIFF
--- a/src/helper.cpp
+++ b/src/helper.cpp
@@ -73,7 +73,7 @@ Color tagColor (const Rules& rules, const std::string& tag)
 ////////////////////////////////////////////////////////////////////////////////
 // Summarize either an active or closed interval, for user feedback.
 std::string intervalSummarize (
-  Database& database,
+  Database&,
   const Rules& rules,
   const Interval& interval)
 {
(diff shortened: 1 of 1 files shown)


src/helper.cpp
```diff
@@ -73,34 +73,15 @@ Color tagColor (const Rules& rules, const std::string& tag)
 ////////////////////////////////////////////////////////////////////////////////
 // Summarize either an active or closed interval, for user feedback.
 std::string intervalSummarize (
-  Database& database,
+  Database&,
```
lauft, line 76: If the parameter is no longer used, it should be removed.

lauft (changes requested):
LGTM, however the unused parameter should be removed from the function.

lauft (approved):
👍🏻

PR #387, opened 2020-09-15
Fixed URL for tarball download

Tronde:
Following the Readme.md to download the tarball in order to build and install timewarrior 1.4.2 I've recognized that the download URL has changed. This PR reflect the change. I've updated the Readme to fit version 1.4.2. Please accept this PR if you like the change.

Maybe you could backport this change to the branch for 1.4.2 as well.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,12 @@ There are two ways to retrieve the Timewarrior sources:
 
 * Or download the tarball with curl,
 
-      curl -O https://taskwarrior.org/download/timew-1.3.0.tar.gz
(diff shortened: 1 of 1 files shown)


Tronde:
I see that the DCO check failed but I'm unsure whether I have to take action here or not.

lauft (approved):
Thanks. 👍🏻

lauft:
As your commit only contains "orthographic" fixes, I think we can skip the DCO.

PR #393, opened 2020-10-17
Update dark_blue.theme

srcr:
fixed the description


lauft (approved):

lauft:
Because it is only a minor fix, we can drop the DCO

lauft:
Thanks for your contribution! 👍🏻

PR #398, opened 2020-11-17
test: Thread.isAlive() -> Thread.is_alive()

sruffell:
Python added Thread.is_alive() in 2.6 and removed Thread.isAlive() in
3.9.

This change is needed to run the tests with Python 3.9.

https://bugs.python.org/issue37804


lauft (approved):
LGTM

Issue #407, opened 2021-01-27
New DateTime literal "earlier"/"before"


quazgar:
Similarly to "later"/"someday", I suggest an new literal "earlier"/"before" for a time way in the past. See the PR below. (I also created one for libshared, is that necessary?)

lauft:
Currently everything that affects `DatetimeParser` will need a corresponding change in `Datetime` in `libshared`. In the long term, `DatetimeParser` should be transferred to `libshared` and the date parsing should be migrated from `Datetime` to `DatetimeParser` altogether.

tbabej:
While I do not see a huge problem with this proposal, the use case is not fully fleshed out in my mind. @quazgar  Could you draft an example (ideally couple of bash lines) of how this would be used in an useful manner? What modes of operation does it enable?

quazgar:
Maybe I am missing an important bit of the syntax, but so far I cannot see a way to define a time range which begins infinitely long ago. For example if I want to find out when time was spent doing `sometag`:

```sh
timew summary sometag             # for today
timew summary sometag soy - later # since the start of the year until end of time
timew summary sometag earlier     # since the beginning of time
timew summary sometag before to yesterday
```

An alternative way to do this would also be welcome, I could also write the documentation for that.

lauft:
Instead of `timew summary TAG earlier` you could use `timew summary TAG :all`... 🤔

quazgar:
Indeed I did not know about `:all` and added documentation about it to the man pages.

quazgar:
I added a bit more information to the man page to explain this choice.

lauft:
After thinking about the topic a long time (see https://github.com/GothenburgBitFactory/timewarrior/pull/408#issuecomment-859332788) I think it would be better, if Timewarrior would support date range constructs like
```
timew ... until/before <datetime>
timew ... since/after <datetime>
```
with this Timewarrior can be used with
```
timew summary since :yesterday
```
instead of
```
timew summary before to yesterday
```

PR #408, opened 2021-01-27
ENH: New literal DateTime "earlier"/"before" = 1970-01-02

quazgar:
See #407:
Similarly to "later"/"someday", I suggest an new literal "earlier"/"before" for a time way in the past.

### DIFF
--- a/doc/man7/timew-dates.7.in
+++ b/doc/man7/timew-dates.7.in
@@ -58,6 +58,8 @@ In addition to the standard date formats, the following are supported:
   <epoch>                               POSIX time
   later                                 2038-01-18T0:00:00 (Y2K38)
   someday                               2038-01-18T0:00:00 (Y2K38)
+  earlier                               1970-01-02T0:00:00
+  before                                1970-01-02T0:00:00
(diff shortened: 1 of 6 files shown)


lauft:
The use cases for this literal seem limited to me - but this also holds true for their counterparts.
I will give this some more thought...

lauft:
Hi @quazgar!
As said in https://github.com/GothenburgBitFactory/timewarrior/pull/408#issuecomment-768543477 I gave this a (very long) thought.

After first agreeing (mainly because of the symmetry to the existing `later`/`someday`) I came to the conclusion that this is in general not the best way to go. Keywords like `later` or `before` should not be bound to a specific date. 

With the latest libshared, `later` in Timewarrior now resolves to `9999-12-30`, which is better than the `2038-01-18` it was before, but still...

Rather than using a specific date, such keywords should resolve to semi-open ranges.

lauft:
I will transfer your updates to the documentation concerning the `:all` hint to the new Asciidoctor files. Thanks for that! 👍🏻

PR #424, opened 2021-05-17
Cleanup for PR 423

sruffell:
Sorry about this, I made a mistake in #423 where getTracked wouldn't return some intervals that intersect with the filter. I also addressed your notes about the unused variables in test/modify.t

Hopefully this can make it in 1.4.3 before you tag.


lauft:
Thanks for the quick update. This will be included.

PR #429, opened 2021-06-11
Add usage to modify command

patdavid:
This commit adds the usage line from the manpage to the modify
command for quick reference when called without parameters:

```
$ timew modify
Usage: timew modify ( start | end ) <id> <date>
Must specify start|end command to modify. See 'timew help modify'.
```

This makes it faster than looking up the syntax in the manpage and
helps build faster memory of syntax for users.


lauft (changes requested):
Good idea. I have some suggestions for the error messages you may take into consideration...

lauft:
@patdavid Will there be any update from your side? Otherwise I would just merge your work with the suggested changes myself...

patdavid:
Ah, sorry - this slipped out of memory for me. :( (sorry!)

I don't have anything else to add at the moment, although I'm happy to revisit this when I can carve out a little more time and maybe propagate the idea to the rest of the commands for more helpful messages.

PR #437, opened 2021-07-13
Minor AtomicFile cleanup

sruffell:
This change just fixes a few minor things in AtomicFile. Primarily to eliminate unnecessary code and make the impl private in the AtomicFile class.


lauft (approved):
LGTM. Thanks again.

PR #441, opened 2021-07-21
Return report return code.

lospatchos:
Report return codes are currently not returned. This PR passes the RC through.

Continues to throw an exception and to return 255 if no output is produced and rc!=0.


lauft (approved):
LGTM, thanks!

Issue #154, opened 2018-06-07
Tags with names matching reports (like 'day') are not recorded


omnster:
It seems to be impossible to track an interval with a tag that matches name of a report, like 'day', 'week', 'month', 'summary' or some initiall letters of the name of that report.  The example output is 

    $ timew start 'summ'  # or 's' or 'su' or 'sum' or 'summar' etc.
    Tracking 
      Started 2018-06-07T13:49:18
      Current               50:24
      Total               0:01:06

If using with taskwarrior, this looks strange
  
    $ task add pro:s +month +day +week 'summary'
    Created task 708
    $ task 708 start
    Starting task d46a2aee 'summary'.
    Tracking
      Started 2018-06-07T13:53:49
      Current            14:01:52
      Total               0:08:47

I have timewarrior 1.1.1
python 2.7.15rc1
Linux 4.15.0-22-generic # 24-Ubuntu SMP Wed May 16 12:15:17 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

lauft:
Related to #60

sruffell:
Just noting that I ran into this myself. You can see below how I tracked 4 intervals, a-d, but interval d silently was not tagged which caused me some confusion.

```
$ src/timew diagnostics

timew 1.2.0
       Platform: Linux

Compiler:
        Version: 7.4.0
           Caps: +stdc +stdc_hosted +LP64 +c8 +i32 +l64 +vp64 +time_t64
     Compliance: C++11

Build Features
          Built: Nov 25 2019 08:35:49
         Commit: 1d27150
          CMake: 3.10.2
     Build type: Debug

Configuration
  TIMEWARRIORDB: -
            Cfg: /home/sruffell/.timewarrior/timewarrior.cfg (-rw- 15 bytes)
       Database: /home/sruffell/.timewarrior (drwx 4096 bytes)
        $EDITOR: vim
    Color theme: Built-in default
                 00 01 02 03 04 05 06 07 08 09 10 11 12

Extensions
       Location: /home/sruffell/.timewarrior/extensions (drwx 4096 bytes)
               (None)

$ for tag in a b c d ; do src/timew start $tag; done; src/timew summary
Note: 'a' is a new tag.
Tracking a
  Started 2019-11-25T09:28:19
  Current                  19
  Total               0:00:00
Recorded a
  Started 2019-11-25T09:28:19
  Ended                    19
  Total               0:00:00
[...]

Issue #60, opened 2018-02-11
[TI-55] Invalid tag: daily


taskwarrior:
_Amado Martinez on 2016-11-26T23:39:29Z says:_ 

 $ timew start myactivity daily
You cannot overlap intervals. Correct the start/end time, or specify the :adjust hint.

I found this issue using the taskwarrior hook with a Task that had a custom "daily" tag.

I wonder if there are other tag names that are not permitted.

taskwarrior:
Migrated metadata:
```
Created: 2016-11-26T23:39:29Z
Modified: 2017-05-21T17:17:56Z
```

taskwarrior:
_Paul Beckingham on 2016-11-28T16:55:27Z says:_ 

 Good catch!  The used of `daily` as a tag is interpreted as a 24-hour recurrence period, and sets the start time to `now-24h`, which may overlap.  I cannot find a workaround so I'm raising the priority.

To answer your question, any tag that matches a date (tuesday, eom, easter ...) or recurrence period (daily, weekly, annual ...) is going to have the same problem.

taskwarrior:
_Thomas Lauf on 2017-04-12T04:38:25Z says:_ 

 Some ideas come into my mind:
1.  repetitions make no sense with TimeWarrior, add a new type to the lexer and ignore it here
1.  move parsing into the command. Commands like `tag`, `start`, `stop` do not expect durations, `track` does
1.  make parsing more context aware. Terms like `day`, `month`, etc should require an amount specifier (`2 days`, `1 month`, `...`)

taskwarrior:
_Paul Beckingham on 2017-05-21T17:17:45Z says:_ 

 I agree Thomas, I would prefer #3, which we can implement using the rat parser.  Moving to backlog for now.

lightdot:
Just got bitten by this bug!

I suddenly got the `You cannot overlap intervals. Correct the start/end time, or specify the :adjust hint.` error and couldn't get the Taskwarrior tasks to hook into the Timewarrior any longer (or so it seemed).

This had me baffled er... _for a time_, since when looking into the Taskwarrior's task information, everything seemed normal.... Timewarrior reports or raw files, normal... systems's date and time, normal...

I made a backup of my Timewarrior database and did some tests, noticing that some of my tasks got recorded exactly 24 hours in the past. And that eventually brought me here.

Two suggestions:

[...]

PR #456, opened 2021-10-23
Add OR and - support for tags (WIP)

### LINKED ISSUE #64: [TI-59] summary or'd tag filter
_lash on 2017-01-13T23:02:46Z says:_ 

 When supplying tags for timew summary, only entries containing all of the tags specificed are shown. I propose possibility for or'ing them instead/aswell.

CURRENT:

Entry 1 has keyword A, B
Entry 2 has keyword B, C

timew summary A B matches entry 1
timew summary B A matches entry 1
timew summary B matches both
timew summary A C matches neither
timew summary C A matches neither

PROPOSED:

Entry 1 has keyword A, B
Entry 2 has keyword B, C

timew summary A B matches entry 1
timew summary B A matches entry 1
timew summary B matches both
timew summary A C matches both
timew summary C A matches both

### LINKED ISSUE #209: I want to be able to exclude certain tags from reports
I find the combination of Taskwarrior and Timewarrior really great, but there is a minor inconvenience in the way TImewarrior works. In Taskwarrior--and that doesn't seem to be properly documented--one can use exclude filters with something like `-tag` to filter out tasks that have a particular tag. The same functionality doesn't seem to be supported in Timewarrior, which instead treats `-` in `-tag` as part of the tag name.

Use case: I use my desktop computer mainly for work, but from time to time I do some tasks that are not related to work. Those tasks are a relatively low in number and I tag them `private`. I would like to be able to exclude them from the reports. Currently, the only options are to tag all other tasks as `work` and then use `timew summary work` or to manually subtract the time for private tasks from the totals, which doesn't work for reports. It would be easier if I could just do something like `timew month -private`.

### LINKED ISSUE #358: How to query multiple tags?
timew summary tag1 tag2 searches for intervals with both tag1 and tag2 in the same time, but what if I do several things during the day and just want to exclude a certain tag from the summary or include more than one tag into the summary?

limdingwen:
This WIP commit adds "OR" and "-" support to tags. For example:

```
% timew summary work OR school

Wk  Date       Day Tags      Start      End    Time   Total
W42 2021-10-23 Sat school  5:00:00  5:21:05 0:21:05
                   school 15:19:12 15:49:17 0:30:05
                   school 16:30:50 17:07:16 0:36:26
                   work   17:07:16 17:52:42 0:45:26
                   school 21:33:54 21:34:33 0:00:39
                   work   21:34:33        - 0:32:45 2:46:26

                                                    2:46:26
```

```
% timew summary -lazy -school

Wk  Date       Day Tags     Start      End    Time    Total
W42 2021-10-23 Sat sleep  5:21:05 14:10:32 8:49:27
                   body  15:49:17 16:30:50 0:41:33
                   work  17:07:16 17:52:42 0:45:26
                   fun   17:52:42 21:33:54 3:41:12
                   work  21:34:33 22:56:28 1:21:55
                   work  22:56:36 23:09:00 0:12:24 15:31:57

                                                   15:31:57
```

Works with other commands like `timew day` as well, which is great. Several problems:

1. Would probably be better as a hint such as `:or`, so users may still use "OR" as a tag. Also for consistency. I need help with this since I'm not sure how to access the hints, or how to add a new hint. I think the `-` operator is good for negation, though it might cause problems if people try to create tags that start with `-`.
2. I had an implementation of a boolean parser, so I could do `timew summary lazy AND school OR work` for example, but that would require `filter._tags` to be a vector instead to preserve ordering, which seems like a big change, so that's not in this commit.
3. Seems to fail 2 tests. `modify.t: timew modify should handle moving start times within an exclusion.` (which states that you cannot overlap intervals) and `stop.t: Add one interval that encloses an exclusion with day change` (which gets a SIGABRT). Not sure how to solve this.

[Original comment on issue](https://github.com/GothenburgBitFactory/timewarrior/issues/358#issuecomment-950158595)

Would close #209.
Would close #64.
Would close #358.

### DIFF
--- a/src/data.cpp
+++ b/src/data.cpp
@@ -451,7 +451,7 @@ bool matchesFilter (const Interval& interval, const Interval& filter)
     for (auto& tag : filter.tags ())
     {
       if (tag == "OR")
-	  {
+      {
         isOr = true;
         break;
       }
@@ -465,11 +465,11 @@ bool matchesFilter (const Interval& interval, const Interval& filter)
       if (isNegative) hasTag = !hasTag;
 
       if (isOr && hasTag)
-	  {
+      {
         return true;
       }
       if (!isOr && !hasTag)
-	  {
+      {
         return false;
       }
     }
--- a/src/data.cpp
+++ b/src/data.cpp
@@ -447,14 +447,30 @@ bool matchesFilter (const Interval& interval, const Interval& filter)
 {
   if (matchesRange (interval, filter))
   {
+    bool isOr = false;
     for (auto& tag : filter.tags ())
     {
-      if (! interval.hasTag (tag))
-      {
+      if (tag == "OR") {
+        isOr = true;
+        break;
+      }
+    }
+
+    for (auto& tag : filter.tags ())
+    {
+      bool isNegative = tag[0] == '-';
+      std::string tagName = isNegative ? tag.substr(1) : tag;
+      bool hasTag = interval.hasTag(tagName);
+      if (isNegative) hasTag = !hasTag;
+
+      if (isOr && hasTag) {
+        return true;
+      }
+      if (!isOr && !hasTag) {
         return false;
       }
     }
-    return true;
+    return isOr ? false : true;
   }
 
   return false;
--- a/src/data.cpp
+++ b/src/data.cpp
@@ -450,7 +450,8 @@ bool matchesFilter (const Interval& interval, const Interval& filter)
     bool isOr = false;
     for (auto& tag : filter.tags ())
     {
-      if (tag == "OR") {
+      if (tag == "OR")
+	  {
         isOr = true;
         break;
       }
@@ -463,10 +464,12 @@ bool matchesFilter (const Interval& interval, const Interval& filter)
       bool hasTag = interval.hasTag(tagName);
       if (isNegative) hasTag = !hasTag;
 
-      if (isOr && hasTag) {
+      if (isOr && hasTag)
+	  {
         return true;
       }
-      if (!isOr && !hasTag) {
+      if (!isOr && !hasTag)
+	  {
         return false;
       }
     }


smemsh:
Keep in mind also, one of the nice features of timewarrior is that tags are more general strings, this is a thing taskwarrior can't really do.  So if we start using `-` and `OR`, this can be a slippery slope.  I'm using tags to store some properties of time intervals with the first character of the tag, which are typically punctuation chars, such as using `+` char to indicate a taskwarrior-mirrored tag, for example.  (Maybe I could store this in the description though, which I currently don't use).

limdingwen:
@smemsh Agreed, and thanks for the anecdote about "-tag" usage. While I'm not sure what would be a good way to do NOTs, I think ORs could be done as a sort of hint (`:or`).

@lauft Thanks for the warm welcome! Honestly, most of that stuff is going over my head right now since I'm new to the codebase, so I'll leave it for the others to discuss. If that refactoring is going to take a long time to make, I think it would help others to add the OR/NOT functionality first, without the boolean parser, since that seems to be most of what was requested for now in the referenced issues.

tbabej:
I might be biased, but long term I'd actually like the CLI syntax for `timew` and `task` grow closer, rather than grow apart. Taskwarrior already has support for things like "or" and "-" (exclusion), so it would make sense to adopt same syntactical elements in `timew`, and extend the tag specification (as @smemsh points out) in `task`. Then we could get these features basically for free, with a consistent UX for the users.

lauft:
Hello @limdingwen, a small update from my side:

I have pushed branch `refactor/filtering` with a first version of the new filtering. See `CmdSummary` for a first application. Converting the other commands should be easy.

> ... I think it would help others to add the OR/NOT functionality first, without the boolean parser ...

Well, your implementation already contains a boolean parser, though a simple one. 😉 

The important thing is that we now can do the parsing (and filter construction) at the right place: close to the CLI. Whether we can integrate it completely into the CLI (and how) is up to the implementation.

I agree with @tbabej to align the CLI syntax, but we will have to address the issues addressed by @smemsh of using "or" as a tag, and having tags starting with `-` (or `+`, or...). (A similar issue to the former is the conflict of tags which are misinterpreted as dates, see #154, #60).

Maybe quoting could help here:
```
-"or" or "-FOO"
```
meaning matching all intervals not having tag `or` or having tag `-FOO`. 🤔 

Full support with parantheses will most likely require adaptions to parsing itself as one cannot simply add a `(`/`)` into the command line – which in turn means we need to use quoting and then deconstruct the quoted tokens...

sruffell:
> I agree with @tbabej to align the CLI syntax, but we will have to address the issues addressed by @smemsh of using "or" as a tag, and having tags starting with - (or +, or...)...

Just thinking out loud, but I wonder if #363 could help here. Then we add additional hints to give the parser clues.

For example, the following two lines could be equivalent:
```
# Produce a summary without intervals tagged "FOO"
timew summary -FOO
timew summary :not :tag FOO
```

Which would make things like the following work:
```
# Produce a summary without intervals tagged "-FOO"
timew summary :not :tag -FOO
```

lauft:
@sruffell This would be a solution, but then we would definitely deviate from the current Taskwarrior CLI syntax. @tbabej what is your opinion about that?

lauft:
I have updated branch [`refactor/filtering`](https://github.com/GothenburgBitFactory/timewarrior/tree/refactor/filtering): Now `getTracked(...)` and `getIntervalsById(...)` use the new filtering – which also simplifies the DB interface a bit. Maybe also `getLatestInterval(...)` can also be replaced (by something like `first_of (all_in_range ({ 0, 0 }))`)... 🤔 

@limdingwen Do you think you can adapt your first version of the new tag filtering to this? Maybe at first only for `CmdSummary`, later for all others where applicable...

Thinking about the new tag filtering: a feature flag may be appropriate here, so we do not surprise any user.

lauft:
@limdingwen I have done some thinking about the filtering, here are my findings so far:

There are three classes of filters:
* _Id_ filters 
* _Tag_ filters
* _Range_ filters 

An _Id filter_ excludes all other filters in the execution. If an _id filter_ is used, only a range and/or a tag set are expected on the CLI. This does not mean a command cannot work with both, but it cannot use both at the same time (see `continue`).

I have compiled an overview what filters are used in the different commands in the following table (_id_ means the command accepts only a single id, _id+_ meaning at least 2):

| command     | id    | id+   | tags  | range |
|    :---:    | :---: | :---: | :---: | :---: |
| modify      |  ✅  |  ❌  |  ❌  |  ❌  |
| move        |  ✅  |  ❌  |  ❌  |  ❌  |
| annotate    |  ✅  |  ✅  |  ❌  |  ❌  |
| delete      |  ✅  |  ✅  |  ❌  |  ❌  |
| fill        |  ✅  |  ✅  |  ❌  |  ❌  |
| lengthen    |  ✅  |  ✅  |  ❌  |  ❌  |
| resize      |  ✅  |  ✅  |  ❌  |  ❌  |
| shorten     |  ✅  |  ✅  |  ❌  |  ❌  |
| split       |  ✅  |  ✅  |  ❌  |  ❌  |
| tag         |  ✅  |  ✅  |  ❌  |  ❌  |
| untag       |  ✅  |  ✅  |  ❌  |  ❌  |
| chart       |  ❌  |  ❌  |  ✅  |  ✅  |
| export      |  ❌  |  ❌  |  ✅  |  ✅  |
| get         |  ❌  |  ❌  |  ✅  |  ✅  |
| report      |  ❌  |  ❌  |  ✅  |  ✅  |
| summary     |  ❌  |  ❌  |  ✅  |  ✅  |
| tags        |  ❌  |  ❌  |  ✅  |  ✅  |
| gaps        |  ❌  |  ❌  |  ✅  |  ✅  |
| continue    |  ✅  |  ❌  |  ✅  |  ❌  |
| join        |  ❌  |  ✅  |  ❌  |  ❌  |
| cancel      |  ❌  |  ❌  |  ❌  |  ❌  |
| config      |  ❌  |  ❌  |  ❌  |  ❌  |
| default     |  ❌  |  ❌  |  ❌  |  ❌  |
| diagnostics |  ❌  |  ❌  |  ❌  |  ❌  |
| extensions  |  ❌  |  ❌  |  ❌  |  ❌  |
| help        |  ❌  |  ❌  |  ❌  |  ❌  |
| show        |  ❌  |  ❌  |  ❌  |  ❌  |
| start       |  ❌  |  ❌  |  ❌  |  ❌  |
| stop        |  ❌  |  ❌  |  ⭕️  |  ❌  |
| track       |  ❌  |  ❌  |  ❌  |  ❌  |
| undo        |  ❌  |  ❌  |  ❌  |  ❌  |

Most of the commands accept either and _id filter_ or a _tag and range_ filter - but there are some outliers:

* The `continue` command allows either a _single id_ or _tag_ filter, but no range filtering. The range is used to define the range of the interval to be created:
  ```
  $ timew continue @1  <from> - <to>  # create new interval <from> - <to> based on @1
  $ timew continue TAG <from> - <to>  # create new interval <from> - <to> based on first find with TAG
  ```
  Here, the error case of supplying an id and tag(s) on the CLI must be handled.

* The `join` command, in contrast to the others with _id+_ (which also accept a single id), only works with two ids.
  This means supplying only one id must result in an error message.

* The `stop` command uses the tag filter not to search for intervals but to filter the tag set of the current open interval to determine whether to create a new open interval:
  ```
  $ timew start FOO BAR BAZ
  $ timew stop  BAR BAZ     # creates closed interval 'FOO' 'BAR' 'BAZ', new open with 'FOO'
  ```
  So this might also be desired (but likely out of scope):
  ```
  $ timew start FOO BAR BAZ
  $ timew stop  -FOO        # creates closed interval 'FOO' 'BAR' 'BAZ', new open with 'FOO'
  ```

limdingwen:
Hi @lauft, for now I've simply copy-pasted my basic OR/- code (without boolean parsing) into `IntervalFilterAllWithTags`. I'm glad to say it works rather well. This isn't intended to make it to release, but it shows the refactor is working with these changes. 😃 

Screenshot 2021-11-11 at 12 13 56 AM

Based on my understanding, this current version would not support boolean parsing yet due to `CLI::getFilter(Range)` returning an Interval which does not contain ordered information, and more work needs to be done there. Is it intended such that `CLI::getFilter(Range)` will return a `IntervalFilter` instead to aid in this?

Also -- apologies, I may have borked this PR :( Still new to GitHub forking, deleted my fork to somehow get the new branch into my repo, and now GitHub doesn't seem to detect changes in my `tag-operators` fork. I'll have to close this PR and open a new one; would make sense too, since we're changing the base over to the refactoring branch instead.

New PR based on refactoring branch is here: #462

PR #458, opened 2021-11-03
Add man documentation for configuring tag colors.

limdingwen:
Adds this into `man timew-tags`:

```
CONFIGURATION
       tags.<tag>.color
           Assigns a specific foreground and background color to a tag, instead of the default color palette determined by your current theme.
           Examples of valid colors include 'white', 'gray8', 'black on yellow', and 'rgb345'.
```

### DIFF
--- a/doc/man1/timew-tags.1.adoc
+++ b/doc/man1/timew-tags.1.adoc
@@ -14,5 +14,5 @@ When a filter is specified, shows only the tags that were used during that time.
 == CONFIGURATION
 
 **tags.**__<tag>__**.color**::
-Assigns a specific foreground and background color to a tag, instead of the default color palette determined by your current theme. +
+Assigns a specific foreground and background color to a tag.
 Examples of valid colors include 'white', 'gray8', 'black on yellow', and 'rgb345'.
--- a/doc/man1/timew-tags.1.adoc
+++ b/doc/man1/timew-tags.1.adoc
@@ -14,5 +14,5 @@ When a filter is specified, shows only the tags that were used during that time.
 == CONFIGURATION
 
 **tags.**__<tag>__**.color**::
-Assigns a specific foreground and background color to a tag, instead of the default color palette determined by your current theme.
+Assigns a specific foreground and background color to a tag, instead of the default color palette determined by your current theme. +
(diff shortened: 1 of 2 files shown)


doc/man1/timew-tags.1.adoc
```diff
@@ -10,3 +10,9 @@ timew-tags - display a list of tags
 == DESCRIPTION
 Displays all the tags that have been used by default.
 When a filter is specified, shows only the tags that were used during that time.
+
+== CONFIGURATION
+
+**tags.**__<tag>__**.color**::
+Assigns a specific foreground and background color to a tag, instead of the default color palette determined by your current theme. +
```
lauft, line 17: I would use `+` only in rare circumstances. Start a new paragraph instead.

lauft (changes requested):
As this also affects the coloring of tags in the `chart` and `sum` command, the documentation there should also be updated 👉🏻 
* https://github.com/GothenburgBitFactory/timewarrior/blob/dev/doc/man1/timew-chart.1.adoc
* https://github.com/GothenburgBitFactory/timewarrior/blob/dev/doc/man1/timew-summary.1.adoc

Note, the handling of the coloring is slightly different there.

doc/man1/timew-tags.1.adoc
limdingwen: Alright. I opted to remove the linebreak instead to keep it consistent with the other configuration options, since the line lengths seem similar enough.

limdingwen:
Yeah, just did another check on the code, seems like the tags are indeed not getting colored for summary.

On line 79: `  auto tag_colors = createTagColorMap (rules, palette, tracked);`

But it's not used anywhere else in the function, specifically here:

On line 158: `      table.set (row, (ids ? 4 : 3), tags);`

Is this intended?

lauft:
> I removed the mention to the default color palette in tags ... I think that's what you mean by different color handling?

Yes 👍🏻 

> Not sure what you meant by `sum` though. 

I meant the `summary` command. But as you found out, `tag_colors` is not used at all – since the beginning... 😂 
I guess we omit this until this got fixed

limdingwen:
Alright! :) Seems like this is ready to be merged then. I’ll create an issue for the tag colouring in summary.

lauft (approved):

PR #462, opened 2021-11-10
WIP: Add complex tag filtering

### LINKED ISSUE #64: [TI-59] summary or'd tag filter

### LINKED ISSUE #209: I want to be able to exclude certain tags from reports

### LINKED ISSUE #358: How to query multiple tags?

limdingwen:
Continuation of #456. This WIP PR aims to add complex tag filtering, allowing for boolean operators to be done on tags. Example:

Screenshot 2021-11-11 at 12 13 56 AM

TODO:
- [ ] Refactor code such that boolean parser is possible
- [ ] Add boolean parser, based off of taskwarrior complex filtering?
- [x] Add basic OR/- support as a test
- [x] Add feature flag
- [ ] Add documentation
- [ ] Add tests

Would close #209.
Would close #64.
Would close #358.

### DIFF
--- a/src/IntervalFilterAllWithTags.cpp
+++ b/src/IntervalFilterAllWithTags.cpp
@@ -32,13 +32,33 @@ IntervalFilterAllWithTags::IntervalFilterAllWithTags (std::set <std::string> tag
 
 bool IntervalFilterAllWithTags::accepts (const Interval& interval)
 {
+  bool isOr = false;
   for (auto& tag : _tags)
   {
-    if (! interval.hasTag (tag))
+    if (tag == "OR")
+    {
+      isOr = true;
+      break;
+    }
+  }
+
+  for (auto& tag : _tags)
+  { 
+    bool isNegative = tag[0] == '-';
+    std::string tagName = isNegative ? tag.substr(1) : tag;
+    bool hasTag = interval.hasTag(tagName);
+    if (isNegative) hasTag = !hasTag;
+
+    if (isOr && hasTag)
+    {
+      return true;
+    }
+
+    if (!isOr && !hasTag)
     {
       return false;
     }
   }
 
-  return true;
+  return isOr ? false : true;
 }
--- a/src/IntervalFilterAllWithTags.cpp
+++ b/src/IntervalFilterAllWithTags.cpp
@@ -27,38 +27,59 @@
 #include <IntervalFilterAllWithTags.h>
 #include <Interval.h>
 
-IntervalFilterAllWithTags::IntervalFilterAllWithTags (std::set <std::string> tags): _tags (std::move(tags))
+IntervalFilterAllWithTags::IntervalFilterAllWithTags (std::set <std::string> tags, const bool complexFiltering): _tags (std::move(tags)), _complexFiltering (complexFiltering)
 {}
 
 bool IntervalFilterAllWithTags::accepts (const Interval& interval)
 {
-  bool isOr = false;
-  for (auto& tag : _tags)
-  {
-    if (tag == "OR")
+  if (_complexFiltering) {
+    // TODO: WIP, This does not fully implement boolean logic.
+    // Check if OR exists.
+    bool isOr = false;
+    for (auto& tag : _tags)
     {
-      isOr = true;
-      break;
+      if (tag == "OR")
+      {
+        isOr = true;
+        break;
+      }
     }
+
+    for (auto& tag : _tags)
+    {
+      // Process negation
+      bool isNegative = tag[0] == '-';
+      std::string tagName = isNegative ? tag.substr(1) : tag;
+      bool hasTag = interval.hasTag(tagName);
+      if (isNegative) hasTag = !hasTag;
+
+      // Return true at the first sign of a tag, if we're using OR logic.
+      if (isOr && hasTag)
+      {
+        return true;
+      }
+
+      // Return false at the first sign of miss, if we're using AND logic.
+      if (!isOr && !hasTag)
+      {
+        return false;
+      }
+    }
+
+    // OR logic : Nothing triggered the match, so this interval does not match.
+    // AND logic: Nothing triggered the miss, so this interval matches.
+    return isOr ? false : true;
   }
-
-  for (auto& tag : _tags)
-  { 
-    bool isNegative = tag[0] == '-';
-    std::string tagName = isNegative ? tag.substr(1) : tag;
-    bool hasTag = interval.hasTag(tagName);
-    if (isNegative) hasTag = !hasTag;
-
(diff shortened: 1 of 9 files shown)


limdingwen:
```
src % timew config complexFiltering off     
No changes made.
src % timew summary school OR work :yesterday
No filtered data found in the range 2021-11-10T00:00:00 - 2021-11-11T00:00:00 tagged with OR, school, work.
src % timew config complexFiltering on       
Are you sure you want to change the value of 'complexFiltering' from 'off' to 'on'? (yes/no) yes         
Config file /Users/ignis/.timewarrior/timewarrior.cfg modified.
src % timew summary school OR work :yesterday

Wk  Date       Day Tags      Start      End    Time    Total
W45 2021-11-10 Wed work    0:13:55  0:56:52 0:42:57
                   work   10:37:59 12:04:13 1:26:14
                   school 12:34:33 13:39:19 1:04:46
                   [snip]
```

Feature flag added under `complexFiltering`.

src/commands/CmdChart.cpp
```diff
@@ -100,7 +100,7 @@ int renderChart (
   // Load the data.
   auto filtering = IntervalFilterAndGroup ({
     new IntervalFilterAllInRange ({ filter.start, filter.end }),
-    new IntervalFilterAllWithTags (filter.tags())
+    new IntervalFilterAllWithTags (filter.tags(), rules.getBoolean("complexFiltering"))
```
lauft, line 103: I would not make this a top-level configuration (like e.g. `verbose`), but put it into a `commands` subgroup. I propose `commands.filtering` with values `simple` (= default) and `complex` (or maybe better: `boolean`).

src/IntervalFilterAllWithTags.cpp
```diff
@@ -27,18 +27,59 @@
...
-IntervalFilterAllWithTags::IntervalFilterAllWithTags (std::set <std::string> tags): _tags (std::move(tags))
+IntervalFilterAllWithTags::IntervalFilterAllWithTags (std::set <std::string> tags, const bool complexFiltering): _tags (std::move(tags)), _complexFiltering (complexFiltering)
 {}
 
 bool IntervalFilterAllWithTags::accepts (const Interval& interval)
 {
-  for (auto& tag : _tags)
-  {
-    if (! interval.hasTag (tag))
+  if (_complexFiltering) {
```
lauft, line 35: As you have already said, this should not be part of `IntervalFilterAllWithTags`, but rather some function or factory which translates the individual command line arguments into filters or filter groups, like e.g. `IntervalFilterOrGroup`, `IntervalFilterWithoutTagSet`, etc.

src/IntervalFilterAllWithTags.cpp
```diff
@@ -27,18 +27,59 @@
...
+    // TODO: WIP, This does not fully implement boolean logic.
+    // Check if OR exists.
+    bool isOr = false;
+    for (auto& tag : _tags)
     {
-      return false;
+      if (tag == "OR")
```
lauft, line 41: Currently, I would opt for lowercase, i.e. `or`

lauft (commented):
Nice POC! 👍🏻 See also the comments in the code.

lauft:
Thanks for the update, @limdingwen !

> Based on my understanding, this current version would not support boolean parsing yet due to CLI::getFilter(Range) returning an Interval which does not contain ordered information, and more work needs to be done there. Is it intended such that CLI::getFilter(Range) will return a IntervalFilter instead to aid in this?

This is not just a simple replacement of `CLI::getFilter(...)`, there are also the error cases (e.g. missing id, wrong number of ids) and their output which have to be handled correctly. Moving the parsing into a function of `CLI` is a goal, but only the final step. So at first, I would implement the construction of the filtering within each command, even if this means code duplication at first.

Therefore as a first step, I would split `CLI::getFilter(<default range>) : Interval` at first into `CLI::getRange(<default range>) : Range` and `CLI::getTagSet() : std::set<std::string>`. From there we can then adapt each command for the new filtering.

Depending on the respective command (see table in [this comment](https://github.com/GothenburgBitFactory/timewarrior/pull/456#issuecomment-964854694)) we can then compose the appropriate filtering. For the new filtering another `CLI`-function is required, returning a vector of strings from which we then can create the complex tag filter.

Later all this could be combined into some kind of `IntervalFilterFactory` (and maybe moved to `CLI`), but I would wait with this until we have a full overview how the implementation in each command looks like.

sruffell:
@lauft, @limdingwen : I might have missed it, but I was wondering if the [refactor/filtering](https://github.com/GothenburgBitFactory/timewarrior/tree/refactor/filtering) branch is going to be submitted via a separate PR?

lauft:
@sruffell I am currently doing some performance tests to check whether something deterioated with `refactor/filtering`. If I spot no errors, I will merge it to dev.

limdingwen:
Hey, just an update -- currently doing finals and some other personal projects, would come back to this when I have the time 😄

lauft:
@limdingwen FYI: I have merged the refactoring of interval filtering to `dev`. I'd be happy to see you to pursue this WIP further.

Good luck for your finals! 🤞🏻

PR #469, opened 2021-12-03
Do not leak IntervalFilters in IntervalFilterAndGroup

sruffell:
This eliminates the use of the naked "new" so that when the filters object is
destroyed, the filters are also freed.

This also eliminates a potential issue with binding an r-value reference to a
Range as a reference in IntervalFilterAllInRange.


lauft (approved):

lauft:
LGTM. Thanks for the update! 👍🏻

PR #470, opened 2021-12-04
Do not leak filter in IntervalFilterFirstOf

sruffell:
This is a continuation of 285aa7171eaf9f5446514ba7, and should have been part
of that commit.

### DIFF
--- a/src/IntervalFilterFirstOf.cpp
+++ b/src/IntervalFilterFirstOf.cpp
@@ -26,7 +26,7 @@
 
 #include <IntervalFilterFirstOf.h>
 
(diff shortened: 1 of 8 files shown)


sruffell:
Sorry I didn't have this one in the previous PR. :/

lauft (approved):

lauft:
> Sorry I didn't have this one in the previous PR. :/

No problem, thanks again for the update! ❤️ 

> Also, what do you think about an IntervalFilterFirst that just accepts the first interval passed to it?

Hm, but there is already IntervalFilterFirstOf...? or what should be the difference between those two?

Issue #338, opened 2020-06-22
Ability to bulk modify intervals based on filter à la Taskwarrior's "modify"


CNG:
Now that I've been using Timewarrior a while, I'm starting to better categorize items (mostly via changing the tags and possibly using annotations as well). I want to do various things like:

* for all intervals matching certain tags, add some other specific tags
* for all intervals with tags matching a certain pattern or regex, add some other specific tags
* rename a tag for all intervals having that tag
* remove a tag from all intervals having that tag

Currently I think I need to do a summary listing of the filter in question, extract all the IDs, and then construct a new command to add or remove tags for all IDs from the previous analysis. It would be GREAT if I could instead do this automatically based on a search, similar to how the `modify` command works in [Taskwarrior - Modify Command](https://taskwarrior.org/docs/commands/modify.html):

> The same change, but to a set of tasks matching the filter:
> 
> `$ task \( project:outdoors or /planting/ \) modify -home +garden`
> 
> Correcting a spelling error in a task description:
> 
> `$ task 1 modify /teh/the/`

I realize Timewarrior currently has a [modify](https://timewarrior.net/docs/timew-modify.1.html) command that unfortunately acts on one interval and the existence of which probably complicates this feature request.

lauft:
> for all intervals matching certain tags, add some other specific tags

The following will work with [1.4.0](https://github.com/GothenburgBitFactory/timewarrior/milestone/10) (see #126):
```
$> timew tag $( timew get dom.tracked.id MATCH_TAG... ) ADDITIONAL_TAG...
```

> for all intervals with tags matching a certain pattern or regex, add some other specific tags
> rename a tag for all intervals having that tag
> remove a tag from all intervals having that tag

Maybe see #179, where a redesign of tagging commands is proposed. As before this works by referencing the target interval via their id(s). Doing this with tags would require to work out where `MATCHING_TAG...` ends in the command - this could prove tricky...

CNG:
Thank you! I think I figured this out, but did that one not get documented, or was I looking in the wrong place?

I was finally coming to test this, and got an error upon running the suggested command (on v1.4.2):
```
$ timew get dom.tracked.id meeting
DOM reference 'dom.tracked.id' is not valid.
```
[...]

PR #487, opened 2022-03-29
Let on-modify.timewarrior rename older intervals too

sskras:
Until now the hook renamed only the current interval, the @1 upon
renaming the task.  With this change it tries to find all existing
intervals with the same Description and rename them all.

Relates to #338

Signed-off-by: Saulius Krasuckas <saulius2@ar-fi.lt>


lauft:
Migrated to https://github.com/GothenburgBitFactory/task-timewarrior-hook/pull/10

PR #500, opened 2022-08-08
Update libshared to 47c3262fa97c4b69542040d39be6c516c38d0e57

### LINKED ISSUE #465: Support for task files in Google Drive and similar mounted volumes
In MacOS, taskwarrior fails if the task data is located in a Google Drive folder. This seems to have been caused by a feature which is not supported by Google Drive - F_FULLSYNC in https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html

(Used in https://github.com/GothenburgBitFactory/libshared/blob/8baf2dbcad9ea24bb9f00a20d912df5da9b45113/src/FS.cpp#L507 )
. Commenting out the code seems to make it work.

Although this can lead to race conditions, using taskwarrior with Google Drive for single person (maybe 2 machines) might not lead to a lot of issues. Can we make this configurable in some way or handle gracefully with warnings?

sruffell:
Needed to allow database to be kept on some network filesystems
(GothenburgBitFactory/libshared#70).

fixes #465


sruffell:
@lauft any reason not to pull this in. The test failures don't *appear* to be related to this change.

lauft:
> any reason not to pull this in?

@sruffell no, except for being a bit swamped. But I am catching up... 😄 
Thanks for the PR.

PR #511, opened 2022-10-28
allow specification of intervals to `export` command

### LINKED ISSUE #510: Export intervals by IDs
It would be useful if we could `timew export @1 @2 @3`, to get just intervals 1, 2, and 3.  In many cases we would already know the ids we want.  Taskwarrior already allows export by ID, which is convenient.  Right now Timewarrior only accepts ranges and tags as filters for `time export`.

smemsh:
This adds feature to specify intervals, ie: `timew export @1 @2 @3`

If intervals are given, `export` will output only those numbered IDs in the JSON.  This allows user that already knows what IDs they want to request only those.  This capability is already available in Taskwarrior export.

Specifying both intervals and either tags or a range, is not allowed.  If the user did not already know what IDs they wanted they would be unable to give them, so it's probably not needed to allow both.  If there is a real scenario where ID set needs to be further narrowed, I'm not sure, so did not try to add that.

Fixes #510

### DIFF
--- a/doc/man1/timew-export.1.adoc
+++ b/doc/man1/timew-export.1.adoc
@@ -5,13 +5,13 @@ timew-export - export tracked time in JSON
 
 == SYNOPSIS
 [verse]
-*timew export* [_<range>_] [_<tag>_**...**]
+*timew export* [_<id>_**...**] | [[_<range>_] [_<tag>_**...**]]
 
 == DESCRIPTION
-Exports all the tracked time in JSON format.
-Supports filtering.
+Exports all the tracked time in JSON format.  Supply either a list of interval IDs (e.g. `@1 @2`), or optional filters (see **timew-ranges(7)** and/or **timew-tags(1)**)
 
 == EXAMPLES
 For example:
 
     $ timew export from 2016-01-01 for 3wks tag1
+    $ timew export @1 @3 @7
--- a/src/commands/CmdExport.cpp
+++ b/src/commands/CmdExport.cpp
@@ -28,6 +28,7 @@
 #include <timew.h>
 #include <iostream>
 #include <IntervalFilterAllInRange.h>
+#include <IntervalFilterAllWithIds.h>
 #include <IntervalFilterAllWithTags.h>
 #include <IntervalFilterAndGroup.h>
 
@@ -38,14 +39,30 @@ int CmdExport (
   Database& database)
 {
   auto filter = cli.getFilter ();
+  std::set <int> ids = cli.getIds ();
+  std::shared_ptr <IntervalFilter> filtering;
 
-  IntervalFilterAndGroup filtering ({
-    std::make_shared <IntervalFilterAllInRange> ( Range { filter.start, filter.end }),
-    std::make_shared <IntervalFilterAllWithTags> (filter.tags())
-  });
+  if (!ids.empty ())
+  {
+    if (!filter.empty ())
+    {
+      throw std::string ("You cannot specify both id and tags/range to export intervals.");
+    }
+    filtering = std::make_shared <IntervalFilterAllWithIds> (ids);
+  }
+  else
+  {
+    filtering = std::make_shared <IntervalFilterAndGroup> (
+      std::vector <std::shared_ptr <IntervalFilter>> (
+        {
+          std::make_shared <IntervalFilterAllInRange> (Range{filter.start, filter.end}),
+          std::make_shared <IntervalFilterAllWithTags> (filter.tags ()),
+        }
+      )
+    );
+  }
 
-  auto intervals = getTracked (database, rules, filtering);
-
+  auto intervals = getTracked (database, rules, *filtering);
   std::cout << jsonFromIntervals (intervals);
 
   return 0;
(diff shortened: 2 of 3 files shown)


smemsh:
I'll write tests and docs later, in meantime looking for feedback

P.S. I could not figure out a way to declare `filtering` at larger scope to factor out to single `getTracked()` call, since the type can be different, but I'm sure there is a way.  I don't know C++ sorry.

lauft (commented):
@smemsh thanks for the PR, I have added my comments.

> P.S. I could not figure out a way to declare filtering at larger scope to factor out to single getTracked() call, since the type can be different, but I'm sure there is a way. I don't know C++ sorry.

I managed it by declaring 
```
std::shared_ptr <IntervalFilter> filtering;
```
before the if-statement and then assigning the respective filters within:
For id filtering:
```
filtering = std::make_shared <IntervalFilterAllWithIds> (ids);
```
The one for the range/tag set proved a bit trickier:
```
filtering = std::make_shared <IntervalFilterAndGroup> (
  std::vector <std::shared_ptr <IntervalFilter>> (
    {
      std::make_shared <IntervalFilterAllInRange> (Range{filter.start, filter.end}),
      std::make_shared <IntervalFilterAllWithTags> (filter.tags ()),
    }
  )
);
```
Finally, one has to dereference the pointer in the call of `getTracked`:
```
auto intervals = getTracked (database, rules, *filtering);
```
I guess there is room for improvement... 🤔

src/commands/CmdExport.cpp
```diff
@@ -38,13 +39,26 @@ int CmdExport (
...
-    std::make_shared <IntervalFilterAllInRange> ( Range { filter.start, filter.end }),
-    std::make_shared <IntervalFilterAllWithTags> (filter.tags())
-  });
-
-  auto intervals = getTracked (database, rules, filtering);
+  if (!ids.empty ())
+  {
+    if (!filter.tags().empty () || filter.start != 0)
+    {
+      throw std::string ("Ids must be specified alone.");
```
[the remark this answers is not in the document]
smemsh: I like the more concise phrasing, but that works too; done.

src/commands/CmdExport.cpp
```diff
@@ -38,13 +39,26 @@ int CmdExport (
...
 
-  IntervalFilterAndGroup filtering ({
-    std::make_shared <IntervalFilterAllInRange> ( Range { filter.start, filter.end }),
-    std::make_shared <IntervalFilterAllWithTags> (filter.tags())
-  });
-
-  auto intervals = getTracked (database, rules, filtering);
+  if (!ids.empty ())
+  {
+    if (!filter.tags().empty () || filter.start != 0)
```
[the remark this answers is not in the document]
smemsh: Oh nice, I was not sure if checking for start time 0 was the right way to even detect there was no range.  `.empty()` which means both, is much better thanks.

lauft:
@smemsh thanks for the update. One commit is fine.
> 1.5.0 is not imminent, right?

1.5.0 is not going to happen next week (I have some stuff on my list I want to get done before that), but I plan to have a release this year. If I had to choose between docs and test, I would opt for docs here.

smemsh:
I made tentative docs, feel free to change it when you merge.  I'll try to get to tests but feel free to merge without it if you release.

doc/man1/timew-export.1.adoc
```diff
@@ -5,13 +5,15 @@ timew-export - export tracked time in JSON
 
 == SYNOPSIS
 [verse]
-*timew export* [_<range>_] [_<tag>_**...**]
+*timew export* [_<interval>_**...**] | [[_<range>_] [_<tag>_**...**]]
```
lauft, line 8: Should be `id` not `interval`

doc/man1/timew-export.1.adoc
```diff
@@ -5,13 +5,15 @@ timew-export - export tracked time in JSON
...
 
 == DESCRIPTION
-Exports all the tracked time in JSON format.
-Supports filtering.
+Exports all the tracked time in JSON format.  Supply either a list of
```
lauft, line 11: One sentence, one line

doc/man1/timew-export.1.adoc
```diff
@@ -5,13 +5,15 @@ timew-export - export tracked time in JSON
...
+intervals (e.g. `@1 @2`), or optional filters (see **timew-ranges(7)** and/or
+**timew-tags(1)**)
 
 == EXAMPLES
 For example:
```
lauft, line 16: Add a description for each example

lauft (approved):
@smemsh Thanks for the update. ❤️ 
I have left some comments (mostly as note to myself, but you can consider them if you supply updates to the documentation).

doc/man1/timew-export.1.adoc
smemsh: didn't add this, it didn't already have description for single existing example, and I feel both examples are self-evident, it's subjective though.

smemsh:
added a simple test, hopefully suitable

PR #529, opened 2023-04-03
Fix summary truncating multibyte characters in long anotations

### LINKED ISSUE #485: Annotation cuts off too early and breaks characters.
If you run `tw summary :ids :annot` it outputs a short version of the annoation which seem's to break specific characters.

Bildschirmfoto 2022-03-10 um 17 47 43

Example characters: ü ö ä
I have seen this happen before when cutoff is done by a byte length which doesn't play well with multi byte characters.

OS: macOS 12.2.1 21D62
Terminal: iTerm2 3.4.9
Shell: zsh 5.8.1 (x86_64-apple-darwin21.3.0)

Cup0fCoffee:
### Description

When calling `summary` command with `:annotations` hint, if the annotation is longer than a certain length, it's going to be truncated. The issue is that the length and truncation are done on a raw string without considering UTF8 multibyte characters, so there are edge cases when an annotation is truncated in the middle of a character creating an invalid UTF8 string.

Fixes #485.

### Notes

~~The fix is trivial, but~~ I have doubts about the test case - I wrote whatever the first thing which came to my mind. Open to feedback in that regard.

Based on the context, we might want to use `utf8_width`, because we care more about how wide the text appears in the output. However, since we can't (or I don't know how to) split the text by width, but we do it by length of characters, I decided to use `utf8_length`. I can see that `libshared` also implements `utf8_text_{length,width}` functions, however as far as I can tell the main difference it that they can filter out the colors, which I don't think matters in this context.

There is one test that is failing for me locally on the clean `develop` branch, so if CI is failing, it might be due to it.

### DIFF
--- a/src/commands/CmdSummary.cpp
+++ b/src/commands/CmdSummary.cpp
@@ -34,6 +34,7 @@
 #include <IntervalFilterAndGroup.h>
 #include <IntervalFilterAllWithTags.h>
 #include <IntervalFilterAllInRange.h>
+#include <utf8.h>
 
 // Implemented in CmdChart.cpp.
 std::map <Datetime, std::string> createHolidayMap (Rules&, Interval&);
@@ -229,9 +230,9 @@ int CmdSummary (
       {
         auto annotation = track.getAnnotation ();
 
-        if (annotation.length () > 15)
+        if (utf8_length (annotation) > 15)
         {
-          annotation = annotation.substr (0, 12) + "...";
+          annotation = utf8_substr (annotation, 0, 12) + "...";
         }
 
         table.set (row, annotation_col_index, annotation);
--- a/test/summary.t
+++ b/test/summary.t
@@ -352,6 +352,16 @@ W\d{1,2} \d{4}-\d{2}-\d{2} .{3}       ?0:00:00 0:00:00 0:00:00 0:00:00
 [ ]+0:00:00
 """)
 
+    def test_multibyte_char_annotation_truncated(self):
+        """Summary correctly truncates long annotation containing multibyte characters"""
+        # Using a blue heart emoji as an example of a multibyte (4 bytes in
+        # this case) character.
+        long_enough_annotation = "a" + "\N{blue heart}" * 20
+        self.t("track FOO sod - sod")
+        self.t("anno @1 " + long_enough_annotation)
+        code, out, err = self.t("summary :anno")
+        self.assertIn("a" + "\N{blue heart}" * 11 + "...", out)
+
 
 if __name__ == "__main__":
     from simpletap import TAPTestRunner


lauft:
@Cup0fCoffee Thank you for your PR!

Your PR definitely fixes the issue of getting unreadable characters due to truncation and also truncates strings single width multi-byte unicode characters better. See the two screenshots below, showing the output of `summary` before and after your fix.

image

image

(see my comment at the end regarding the uneven truncation)

> I have doubts about the test case - I wrote whatever the first thing which came to my mind. Open to feedback in that regard.

As far as I see it, the test does not cover the case when a multi-byte character is actually split: '💙' is four bytes wide and Timewarrior cuts off after 12. I would suggest to add a single width character at the beginning (e.g. 'a'), so we shift the last '💙' into the cut.

> Based on the context, we might want to use utf8_width, because we care more about how wide the text appears in the output. ...

I agree with that – and yes, it's complicated 😅. I guess Timewarrior will need a dedicated truncate function to take care of this... 🤔

P.S.
> There is one test that is failing for me locally on the clean develop branch, so if CI is failing, it might be due to it.

Can you tell me which one? Does it fail every time or is it just flaky?

lauft (commented):
The fix looks good to me.
The issue of uneven truncation can be dealt with in a separate issue.
However I would add a single width character to the test string to have the truncation actually split a multi-byte character.

Cup0fCoffee:
@lauft Thank you for your feedback! I've changed the test string.

> Can you tell me which one? Does it fail every time or is it just flaky?

```
not ok 5 - help.t: timew help with command should show man page
# FAIL: AssertionError on file /home/maxim/sources/timewarrior/test/help.t line 70 in test_help_with_command_should_show_man_page: 'self.assertEqual(actual, expected)':
#       "TIME[1018 chars]1.4.3-dev                                     [84 chars]1)
# " != "TIME[1018 chars]1.4.3                                         [84 chars]1)
# "
#       Diff is 1363 characters long. Set self.maxDiff to None to see it.
```
I figured out the reason for this test failing. Initially I compiled `1.4.3` version and installed it. When I decided to open this PR, I switch to `dev` and was just running tests via `make test`, without `make && sudo make install`. If I do the install of the dev, the test passes. It's not a huge deal, but I guess would be nice for the test case to take not the globally installed man page, but the one which was built within the repo. Hope my explanation makes sense :blush:

lauft:
@Cup0fCoffee Thanks for the quick update!
Could I ask you to also rebase your branch to the latest state of `develop`? 😬 

> I figured out the reason for this test failing... Hope my explanation makes sense.

Yes, it was more or less what I expected. 👍🏻 
Actually, I thought the tests were configured to use the man pages from the repository ([Those lines in `timew.py`](https://github.com/GothenburgBitFactory/timewarrior/blob/c85e0c01caaadad309e0d87b5fa23a451c89d6f6/test/basetest/timew.py#L74-L78) should configure the `MANPATH` for this), but maybe something has been overlooked... 🤔 

>When I decided to open this PR, I switch to dev ...

Btw. the `dev` branch is obsolete, please use the `develop` branch to base your PRs on (👉🏻 https://github.com/GothenburgBitFactory/timewarrior/discussions/472)

Cup0fCoffee:
@lauft Rebased.

> Actually, I thought the tests were configured to use the man pages from the repository ([Those lines in timew.py](https://github.com/GothenburgBitFactory/timewarrior/blob/c85e0c01caaadad309e0d87b5fa23a451c89d6f6/test/basetest/timew.py#L74-L78) should configure the MANPATH for this), but maybe something has been overlooked...

The man pages are called [via an utility function](https://github.com/GothenburgBitFactory/timewarrior/blob/c85e0c01caaadad309e0d87b5fa23a451c89d6f6/test/help.t#L67), not a method of the `Timew` instance, so the env is not passed.

One way to fix it could be to pass the env from `Timew` instance to the utility function:
```diff
diff --git a/test/help.t b/test/help.t
index 786855dd..46f398cb 100755
--- a/test/help.t
+++ b/test/help.t
@@ -64,7 +64,7 @@ class TestHelp(TestCase):

     def test_help_with_command_should_show_man_page(self):
         """timew help with command should show man page"""
-        _, expected, _ = run_cmd_wait_nofail(["man", "timew-start"])
+        _, expected, _ = run_cmd_wait_nofail(["man", "timew-start"], env=self.t.env)
         _, actual, _ = self.t("help start")

         self.assertEqual(actual, expected)
```

PR #538, opened 2023-05-03
Fix man page build for out-of-source builds

### LINKED ISSUE #461: Man pages not installed with out-of-source build in 1.4.3
Building 1.4.3 from source and installing, I see that if I use an out of source cmake build, the man pages do not get installed properly. They do get installed if I use an in-source build. Didn't have this issue with 1.4.2.

Also, it seems that there is an empty directory named CMakeFiles installed in the `man/man1` and `man/man7` directories, for both out of source and in-source builds.

b-fein:
Fixes #461

The source files for asciidoctor need to be looked up in the `SOURCE_DIR`, and the output for them should be in the `BINARY_DIR`. Since they are identical for in-source builds, the previous version worked as well in those cases.


lauft:
Thanks ❤️

PR #541, opened 2023-05-19
fix(totals.py): end report at current time if a task is being tracked

### LINKED ISSUE #540: reports should end at current time if a task is still running
I was using the `totals.py` report and saw some weird behaviour.
I have a task tracked earlier today.
If I run `timew sum`, it tells me how long this task has been running---an hour or so.
If I run `totals.py` with `:day`, it assumes the end of today to be the end of the task, making it 5 hours longer than what it should be.
If I use `totals.py` with `:week`, it adds days to it until the end of the week.

I don't think this is correct behaviour. Users should be able to generate a report "until now", even if they're currently tracking a task---generating a report could be the task itself?

This seems to be because the reports are always passed `temp.report.end`, and in the code I see that this is set to `filter.end.toISO` if the filter hasn't ended:

https://github.com/GothenburgBitFactory/timewarrior/blob/develop/src/commands/CmdReport.cpp#L149

So the check in totals.py never goes into the else branch here:

https://github.com/GothenburgBitFactory/timewarrior/blob/develop/ext/totals.py#L112

I don't know if this is to be considered a bug in the report command that always passes `temp.report.end`, or if in totals.py where it should just end at the present period if a task is being currently tracked? As a workaround, I've tweaked the script to do the latter.

sanjayankur31:
Fixes #540

### DIFF
--- a/ext/totals.py
+++ b/ext/totals.py
@@ -110,7 +110,8 @@ def calculate_totals(input_stream):
         if report_end_utc is not None:
             j[-1]["end"] = min(report_end_utc, datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)).strftime(DATEFORMAT)
         else:
-            report_end = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_end_utc = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_end = report_end_utc.astimezone(tz=to_zone)
     else:
         if report_end_utc is not None:
             j[-1]["end"] = min(report_end_utc, datetime.datetime.now(tz=from_zone)).strftime(DATEFORMAT)
--- a/ext/totals.py
+++ b/ext/totals.py
@@ -110,7 +110,7 @@ def calculate_totals(input_stream):
             report_end = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
     else:
         if report_end_utc is not None:
-            j[-1]["end"] = report_end_utc.strftime(DATEFORMAT)
+            j[-1]["end"] = min(report_end_utc, datetime.datetime.now(tz=from_zone)).strftime(DATEFORMAT)
         else:
             j[-1]["end"] = datetime.datetime.now(tz=from_zone).strftime(DATEFORMAT)
             report_end = datetime.datetime.now(tz=to_zone)
--- a/ext/totals.py
+++ b/ext/totals.py
@@ -100,6 +100,9 @@ def calculate_totals(input_stream):
     if "start" in j[0]:
         if report_start_utc is not None:
             j[0]["start"] = max(report_start_utc, datetime.datetime.strptime(j[0]["start"], DATEFORMAT).replace(tzinfo=from_zone)).strftime(DATEFORMAT)
+        else:
+            report_start_utc = datetime.datetime.strptime(j[0]["start"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_start = report_start_utc.astimezone(tz=to_zone)
     else:
         return ["Cannot display an past open range"]
 


ext/totals.py
```diff
@@ -109,11 +109,8 @@ def calculate_totals(input_stream):
         else:
             report_end = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
     else:
-        if report_end_utc is not None:
-            j[-1]["end"] = report_end_utc.strftime(DATEFORMAT)
-        else:
-            j[-1]["end"] = datetime.datetime.now(tz=from_zone).strftime(DATEFORMAT)
```
lauft, line 115: You are right that overwriting the end time of an open interval with the report end is wrong, but only if the report end time is *after* the current time.

If one would call `timew totals :yesterday` with an interval started yesterday, your fix would also display the time of today.

So I think we should take the minimum of _now_ and _report.end_ instead.

lauft (changes requested):
Thank you for your contribution.

While your fix works for `:day` and `:week`, I think it does not handle `:yesterday` correctly when if the current open interval starts there (see review comment).

Also, please accept the [DCO](https://github.com/GothenburgBitFactory/timewarrior/blob/develop/DCO) by signing your commits accordingly if you want this PR to be accepted. Thanks!

ext/totals.py
sanjayankur31: that makes sense, updated.

sanjayankur31:
I updated both commits to add a signed off by line. 

Btw, when the report is run without arguments, i still see an error. It was meant to be fixed i https://github.com/GothenburgBitFactory/timewarrior/commit/c80442c8786226b5ed4685c4a32e8b0454eaf69c from the looks of it:

```
$ timew report totals.py
Traceback (most recent call last):
  File "/home/asinha/.timewarrior/extensions/totals.py", line 184, in <module>
    for line in calculate_totals(sys.stdin):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/asinha/.timewarrior/extensions/totals.py", line 146, in calculate_totals
    "Total by Tag, for {:%Y-%m-%d %H:%M:%S} - {:%Y-%m-%d %H:%M:%S}".format(report_start, report_end),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unsupported format string passed to NoneType.__format__
'/home/asinha/.timewarrior/extensions/totals.py' returned 1 without producing output.
```

I think it's because `report_start` and `report_end` can both be `None` if `temp.report.start` and `temp.report.end` aren't included in the config that's passed to the script. New issue for this one (if someone else can confirm it)?

lauft:
The PR looks good to me now. 👍🏻 

> Btw, when the report is run without arguments, i still see an error. ... New issue for this one (if someone else can confirm it)?

I confirm this, but we can also fix it right here. 

The fix is quite simple, just add the following `else` statement after line 102:
```
diff --git a/ext/totals.py b/ext/totals.py
--- a/ext/totals.py	(revision c4bb908bcc021320b502d70b860f5ef853fc7900)
+++ b/ext/totals.py	(revision c6c2d1e0cbdccfd8e95deaff8cd7b95a201a6ca1)
@@ -100,6 +100,9 @@
     if "start" in j[0]:
         if report_start_utc is not None:
             j[0]["start"] = max(report_start_utc, datetime.datetime.strptime(j[0]["start"], DATEFORMAT).replace(tzinfo=from_zone)).strftime(DATEFORMAT)
+        else:
+            report_start_utc = datetime.datetime.strptime(j[0]["start"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_start = report_start_utc.astimezone(tz=to_zone)
     else:
         return ["Cannot display an past open range"]
``` 
The case of `temp.report.end` being empty is already handled, but there is also a small bug in the time conversion. It has to be
```
Subject: [PATCH] Fix conversion from UTC to local zone for report end time
---
diff --git a/ext/totals.py b/ext/totals.py
--- a/ext/totals.py	(revision 84fdf76f9aba22d9badbcbdf92942714cb5e254f)
+++ b/ext/totals.py	(revision c4bb908bcc021320b502d70b860f5ef853fc7900)
@@ -107,7 +107,8 @@
         if report_end_utc is not None:
             j[-1]["end"] = min(report_end_utc, datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)).strftime(DATEFORMAT)
         else:
-            report_end = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_end_utc = datetime.datetime.strptime(j[-1]["end"], DATEFORMAT).replace(tzinfo=from_zone)
+            report_end = report_end_utc.astimezone(tz=to_zone)
     else:
         if report_end_utc is not None:
             j[-1]["end"] = report_end_utc.strftime(DATEFORMAT)
```
(aligning it also to the handling of `temp.report.start`).

lauft (approved):

sanjayankur31:
Thanks, I've also incorporated your two fixes here now.

PR #547, opened 2023-06-19
AtomicFile: Operate on the target of symlinks

sruffell:
Currently, if an AtomicFile is opened on a symlink, the AtomicFile would
end up overwriting the link. This change makes AtomicFiles operate on
the targets of the links and not the links themselves.

**NOTE**: Due to the way Path is implemented in libshared, this change
will not preserve links if the link does not have a valid target.

The `Path::exists()` will always follow links, and therefore is_link() will erroneously
return false: 
https://github.com/GothenburgBitFactory/libshared/blob/master/src/FS.cpp#L204

### DIFF
--- a/test/AtomicFileTest.cpp
+++ b/test/AtomicFileTest.cpp
@@ -33,6 +33,7 @@
 #include <sstream>
 #include <string>
 #include <test.h>
+#include <unistd.h>
 
 #ifdef FIU_ENABLE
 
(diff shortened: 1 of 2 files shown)


test/AtomicFileTest.cpp
```diff
@@ -150,16 +151,47 @@ int fiu_test (UnitTest& t)
...
+  }
+  AtomicFile::finalize_all ();
+
+
+  // After finalizing we should be able read the data that we wrote through the
+  // target
+  std::string read_contents;
+  AtomicFile::read (target, read_contents);
+  t.is(read_contents, write_contents, "AtomicFile writes to symlinks end up in target");
+  t.is (link.is_link (), true, "AtomicFile to link shall maintain symlink");
```
lauft: Isn't the group missing here, compared to the other tests? 🤔 
```suggestion
  t.is(read_contents, write_contents, "AtomicFileTest: AtomicFile writes to symlinks end up in target");
  t.is (link.is_link (), true, "AtomicFileTest: AtomicFile to link shall maintain symlink");
```

sruffell: Good catch, updated.

lauft:
Thanks for the PR! ❤️

Issue #515, opened 2022-12-02
retag command


quazgar:
There should be a `retag` command which should be equivalent to `untag` and `tag` combined:

If interval `@1` is tagged as *foo*, the following should be equivalent:
```console
timew retag @1 bar
```
and
```console
timew tag @1 bar
timew untag @1 foo
```

Basically, `retag` should replace all existing tags with the ones given as the new arguments.

smemsh:
it would be nice, but this works in the meantime:
``` bash
timeredo ()
{
    local n i ival
    local -a ivals olds

    while [[ $1 =~ ^@ ]]; do ivals+=($1); shift; done
    n=${#ivals[@]}
    if ((n == 0 || $# == 0)); then
        echo "$FUNCNAME: overwrite tags for given intervals" >&2
        echo " usage: $FUNCNAME [@interval]... [tag]..." >&2
        false
        return
    fi

    for ((i = 0; i < n; i++)); do
        ival=${ivals[i]#@}
        if (($(timew get dom.tracked.$ival.tag.count))); then
            olds=($(
                timew get dom.tracked.$ival.json |
                jq -r '.tags[]'
            ))
            if ! timew untag @$ival "${olds[@]}"
            then echo "untag failed" >&2; false; return; fi
        fi
        if ! timew tag @$ival "$@"
        then echo "tag failed" >&2; false; return; fi
    done
}
```

quazgar:
I just added a bit more documentation links in #554

PR #551, opened 2023-07-18
Add retag command

cyberme0w:
Based on issue #515, I've added the function to retag intervals. This works by first clearing the interval of any tags, then adding the ones fetched from the args and stored under `tags`.

For instance, given an interval `@1` with the tags 'foo' and 'bar', running `timew retag @1 buz` will result in the interval dropping the previous 'foo' and 'bar' tags and gaining the 'buz' tag.

Glad to help, and please let me know if there's anything I should change or improve.


lauft (approved):
Thanks for your contribution! ❤️ 

The only thing I would change is the copyright date, which should be just for 2023 for the new files, i.e. `CmdRetag.cpp` and `retag.t`

cyberme0w:
Happy to help! I've been using timewarrior for some time now as it's perfect for my workflow, so I decided I might as well try to give back where I can.

lauft (approved):

PR #552, opened 2023-07-25
Extend XDG support to macOS

Rahlir:
I was a little mad about #207 not supporting macOS and the docs stating that macOS is a non-unix system. So I just made a tiny pull request to remedy this. I very much appreciate the work done by the original PR author, so I apologize if my comment on his PR was a bit too strongly worded. I am also not really sure why this feature should be limited to UNIX at all, but I leave this up to the maintainer.


lauft (approved):
Thanks for your contribution! ❤️

PR #607, opened 2024-05-25
Fix manual page formatting

tobiolo:
Remove brackets and custom link style

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -7,8 +7,7 @@ add_subdirectory (themes)
 
 if (ASCIIDOCTOR_FOUND)
   set (ASCIIDOCTOR_OPTIONS "--attribute=manmanual=User Manuals"
-                           "--attribute=mansource=timew ${PROJECT_VERSION}"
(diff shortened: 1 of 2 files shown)


lauft (changes requested):
The asciidoc documents are also used for the HTML documentation on timewarrior.net (👉🏻 https://timewarrior.net/reference/timew.1/).
While the man pages are unchanged, formatting the links like this makes them no longer clickable in HTML.

tobiolo:
> The asciidoc documents are also used for the HTML documentation on timewarrior.net (👉🏻 https://timewarrior.net/reference/timew.1/). While the man pages are unchanged, formatting the links like this makes them no longer clickable in HTML.

Thanks for your feedback. I removed the brackets then and also the custom linkstyle settings for the AsciiDoctor.

Now this looks better for me:
image

lauft (approved):

lauft:
Thank you for your contribution. ❤️

tobiolo:
> Thank you for your contribution. ❤️ 

You are welcome!

PR #632, opened 2024-08-21
Fix py3.12 warnings for datetime usage, non-raw regex strings

smemsh:
After upgrading my `python3` from `python-3.9` to `python-3.12`, running the `test` target causes a bunch of new warnings to be emitted.  These are regarding `datetime.datetime.utcnow()` being deprecated.  This is because upstream wants to make changes to naive datetimes (those without `tzdata`).

Here is the blurb from https://docs.python.org/3/whatsnew/3.12.html:

> datetime: datetime.datetime’s utcnow() and utcfromtimestamp() are deprecated and will be removed in a future version. Instead, use timezone-aware objects to represent datetimes in UTC: respectively, call now() and fromtimestamp() with the tz parameter set to datetime.UTC.

There is some discussion of upstream's rationale in python/cpython#81669.

For the fix, I chose to use `datetime.timezone.utc` rather than `datetime.UTC` because the latter was only added in `python-3.11` and I wanted the tests to work under both `python-3.9` and `python-3.12`.

The following edit sequences were run to produce the changes:

First, `test_totals.t` had imported all of `datetime`, unlike all the other files which were importing individual names `from datetime`.  It also did a wildcard import from `ext/totals.py` which caused a shadowing issue so I made the imports more specific.  Now taht file imports `datetime` like the others, so later mass-edits (below) will work on this file too:
```
sed -i \
-e 's/^.*totals.*\*/from totals import format_seconds, calculate_totals/' \
-e 's/^import.datetime$/from datetime import datetime, timedelta/' \
-e 's/datetime\.datetime\./datetime./g' \
-e 's/datetime\.timedelta/timedelta/' \
test/test_totals.t
```

Next, we replace the deprecated `utcnow()` with a tzinfo-aware call, which can be done directly by giving a zone argument to `now()` but in other uses, a tzinfo-naive object is obtained, used for some things, and then later its `.utcnow()` method is run.  We had to convert those uses in-place by first localizing:
```
grep -lsrF \
  -e 'datetime.now().utcnow()' \
  -e 'now.utcnow()' \
| xargs sed -i \
  -e 's/datetime.now().utcnow()/datetime.now(timezone.utc)/g' \
  -e '/import.datetime/a\' \
  -e 'from dateutil import tz' \
  -e 's/^\(from datetime import datetime,\)/\1 timezone,/' \
  -e 's/now.utcnow()/now.replace(tzinfo=tz.tzlocal()).astimezone(timezone.utc)/'
```

Finally, I removed the `tzlocal` import from files that didn't need it because they didn't do a conversion from naive objects to a aware objects (in the transform above):
```
for file in $(grep -lsr 'from.dateutil.import.tz$' test)
do grep -q tzlocal $file || sed -i '/^from.dateutil.import.tz$/d' $file
done
```

Incidentally, there were some regexes in `summary.t` that are not raw-quoted, generating a warning on embedded `\d` in non-raw quoted strings.  These were probably already warnings before.  They were fixed en passant in a separate commit.

Please apply, thanks.

### DIFF
--- a/test/annotate.t
+++ b/test/annotate.t
@@ -29,7 +29,8 @@
 import os
 import sys
 import unittest
-from datetime import datetime, timedelta
+from datetime import datetime, timezone, timedelta
+from dateutil import tz
 
 # Ensure python finds the local simpletap module
 sys.path.append(os.path.dirname(os.path.abspath(__file__)))
@@ -44,7 +45,7 @@ class TestAnnotate(TestCase):
 
     def test_add_annotation_to_open_interval(self):
         """Add an annotation to an open interval"""
-        now_utc = datetime.now().utcnow()
+        now_utc = datetime.now(timezone.utc)
         one_hour_before_utc = now_utc - timedelta(hours=1)
 
         self.t("start {:%Y-%m-%dT%H:%M:%S}Z".format(one_hour_before_utc))
@@ -58,7 +59,7 @@ class TestAnnotate(TestCase):
 
     def test_should_use_default_on_missing_id_and_active_time_tracking(self):
         """Use open interval when adding annotation with missing id and active time tracking"""
-        now_utc = datetime.now().utcnow()
+        now_utc = datetime.now(timezone.utc)
         one_hour_before_utc = now_utc - timedelta(hours=1)
         two_hours_before_utc = now_utc - timedelta(hours=2)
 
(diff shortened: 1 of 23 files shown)


smemsh:
oops, my description.md was transformed by my sed accidentally, let me fix that...

lauft:
Thanks ❤️

PR #633, opened 2024-08-22
Make week number parsing ISO8601 compliant

### LINKED ISSUE #595: using week numbers accounts Sunday through Saturday, not Monday through Sunday
Week numbers would typically be ISO weeks, but they seem to include the Sunday from the previous week, and not include the Sunday from the given week.  For example:
```
timew summary 2024-W10 - 2024-W11
```
includes Sunday from W9 and does *not* include Sunday from W10.  This differs from [imo correct] behavior of `:week` and `:lastweek` which use Monday based weeks.  I would think week number would correspond exactly to `date +%V`.

This was tested with timew-1.7.1 and its pegged libshared.

Got the week syntax from `src/libshared/test/datetime.t.cpp`.  Was not sure if the bug should be filed in libshared repo, but it seems to be library-user-specific since `:week` and `:lastweek` are working as expected, but let me know if this should be moved there.

smemsh:
This PR makes week number parsing ISO8601-compliant.  Previously, week numbers parsed as Sunday-based when given as `2024-W34`, and did not respect the ISO8601 algorithm for which year they were in.  This was inconsistent with `:week`, `:lastweek`, etc, which use Monday-based weeks.  This PR equalizes that inconsistency, and switches to the ISO standard algorithm for week numbers.

As a side effect, note that day number parsing (ie `2024-W34-1`) also changes from 0-6 Sunday to 1-7 Monday, again following ISO8601.  This only has a consequence for Sunday itself, which changes from day 0 to day 7.

**NOTE:** *These changes strictly duplicate* ones from GothenburgBitFactory/libshared#81.  Timewarrior contains duplicate implementations of several methods from libshared's `Datetime` class in its own `DatetimeParser` class.  All methods we have changed are identical in both places.  Eventually, the two classes are [intended be merged back together](https://github.com/GothenburgBitFactory/libshared/pull/81#issuecomment-2262105316).

The only difference with the libshared version of the patch is that we use libshared's `Datetime::dayOfWeek()` and `Datetime::daysInYear()` because they aren't among the copied methods in our `DatetimeParser` class.

Rationale and details, including references and links to standards, can be seen in the discussion and commits in libshared.

No tests were added to Timewarrior, as all my changes are already fully covered in new libshared tests, and all the existing tests here pass with no changes.

Fixes #595.

### DIFF
--- a/src/DatetimeParser.cpp
+++ b/src/DatetimeParser.cpp
@@ -2787,7 +2787,7 @@ bool DatetimeParser::isOrdinal (const std::string& token, int& ordinal)
 bool DatetimeParser::validate ()
 {
   // _year;
-  if ((_year    && (_year    <   1900 || _year    >                                  2200)) ||
+  if ((_year    && (_year    <   1900 || _year    >                                  9999)) ||
       (_month   && (_month   <      1 || _month   >                                    12)) ||
       (_week    && (_week    <      1 || _week    >                                    53)) ||
       (_weekday && (_weekday <      0 || _weekday >                                     6)) ||
--- a/src/DatetimeParser.cpp
+++ b/src/DatetimeParser.cpp
@@ -190,7 +190,7 @@ void DatetimeParser::clear ()
   _year    = 0;
   _month   = 0;
   _week    = 0;
-  _weekday = 0;
+  _weekday = Datetime::weekstart;
   _julian  = 0;
   _day     = 0;
   _seconds = 0;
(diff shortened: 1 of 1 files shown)


smemsh:
Waiting for GothenburgBitFactory/libshared#81 to get merged first.  I will then update the libshared peg, add to this PR, and undraft so we can merge this one.

smemsh:
> Waiting for GothenburgBitFactory/libshared#81 to get merged first

Thanks @lauft! The libshared peg has been advanced, this one is ready.

lauft:
Thanks ❤️

PR #668, opened 2025-05-26
Add undo to zsh completion

Jaffex:
Added the `undo` command to zsh completion, removed `revert` (which isn't a valid timewarrior command)


lauft:
@Jaffex Thanks for your contribution! ❤️ 

However, the zsh completion is only bundled into Timewarrior, but maintained here 👉🏻 https://github.com/ianmkenney/timewarrior_zsh_completion. Please move your PR to there.

PR #673, opened 2025-06-02
Fix Fish completions

frangio:
Fish Shell completions are broken. Tags are extracted with descriptions and extra surrounding spaces. Additionally, the completions are evaluating when the script loads rather than when requested.

### DIFF
--- a/completion/timew.fish
+++ b/completion/timew.fish
@@ -8,7 +8,10 @@ function __fish_timew_get_commands
 end
 
 function __fish_timew_get_tags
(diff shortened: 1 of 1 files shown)


lauft:
`timew.fish` is included from [pfmephisto/timew-fishcompletion](https://github.com/pfmephisto/timew-fishcompletion) (👉🏻 [README.md](https://github.com/GothenburgBitFactory/timewarrior/blob/develop/completion/README.md))
Please open a pull-request there

PR #693, opened 2025-08-01
Expose data location through DOM

### LINKED ISSUE #692: Expose data location for integration with third-party tools
Since the data location can be one of a few directories, third party tools would need to determine which to use. Take the following, in order of what timewarrior prioritizes when searching for data:

1. `$HOME/.timewarrior/data` (legacy)
2. `$XDG_DATA_HOME/timewarrior/data` (preferred)
3. `$HOME/.local/share/timewarrior/data`

The order by which timewarrior searches for the active data directory is not the same order for which it would create a fresh database if no database was found. This becomes even more confusing given the following behavior:

1. `$HOME/.timewarrior` exists (note no `data/`) and `$HOME/.local/share/timewarrior/data` exists. Using `timew` creates `$HOME/.timewarrior/data` and is now the default with empty data.
2. `$HOME/.local/share/timewarrior/data` exists, but down the line a user defines `$XDG_DATA_HOME` (to something other than `~/.local/share/`). The next run of `timew` prompts the user to create a database as if one wasn't found.
3. `$HOME/.timewarrior/data/` exists, but down the line a user defines `$XDG_DATA_HOME`. The next run of `timew` **does not** prompt the user to create a database.

Because of this, it would be nice to be able to ask timewarrior where the data it will be using is.

See [motivation here](https://github.com/ianmkenney/timewarrior_zsh_completion/pull/6).

ianmkenney:
This PR introduces a new `timew get` target for `dom.data.location` which returns the database path. Closes #692.

### DIFF
--- a/src/Database.cpp
+++ b/src/Database.cpp
@@ -297,6 +297,13 @@ std::string Database::getLatestEntry ()
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+std::string Database::getLocation ()
+{
+  auto datapath = Path (_location);
+  return datapath;
+}
+
+////////////////////////////////////////////////////////////////////////////////
 void Database::addInterval (const Interval& interval, bool verbose)
 {
   assert ((interval.end == 0) || (interval.start <= interval.end));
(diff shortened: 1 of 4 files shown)


ianmkenney:
This does currently return `~/.local/share/timewarrior/data`, which should be expanded...

PR #708, opened 2025-09-03
add retag command to fish completions

funnes:


lauft:
@funnes The Fish completion is managed in https://github.com/pfmephisto/timew-fishcompletion. Please transfer your pull request to there.

PR #717, opened 2025-10-04
Add CodeQL analysis workflow configuration

lauft:


lauft:
Incorporated changes into #716

Issue #591, opened 2024-02-26
Tags with special characters (e.g. `+`, `-`) do not apply tag coloring for `tags.<Tag>.color`


erik-ahlburg-schmidt:
I want to color my tags by project and by use-case. For tags without any special characters this works, but once I have tags which contain special characters like `+` or `-` this is not working. I have projects which use the `+` sign and the `-` sign is used for tickets. 

Here is the config (when running `timew config`):
```
color = on
confirmation = on
debug = off

define exclusions:
  friday = <8:30 12:00-13:00 >17:30
  monday = <8:30 12:00-13:00 >17:30
  saturday = >0:00
  sunday = >0:00
  thursday = <8:30 12:00-13:00 >17:30
  tuesday = <8:30 12:00-13:00 >17:30
  wednesday = <8:30 12:00-13:00 >17:30

journal:
  size = -1

define reports:
  day:
    axis = internal
    cell = 15
    day = yes
    holidays = no
    hours = all
    lines = 2
    month = no
    spacing = 1
    summary = yes
    totals = no
    week = no
    weekday = yes
  month:
    cell = 15
    day = yes
    holidays = yes
    hours = all
    lines = 1
    month = yes
    spacing = 1
    summary = yes
    totals = yes
    week = yes
    weekday = yes
  summary:
    annotations = yes
    holidays = yes
    ids = yes
  week:
    cell = 15
    day = yes
    holidays = yes
    hours = all
    lines = 1
    month = no
    spacing = 1
    summary = yes
    totals = yes
    week = yes
    weekday = yes

define tags:
  XL2:
    color = color09

define theme:
  colors:
    debug = blue
    exclusion = gray8 on gray4
    holiday = gray4
    ids = color13
    label = gray4
    today = rgb521
  description = dark.theme: A default, 256-color theme
  palette:
    color01 = white on red
    color02 = white on blue
    color03 = black on green
    color04 = black on magenta
    color05 = black on cyan
    color06 = black on yellow
    color07 = black on white
    color08 = white on bright red
    color09 = white on bright blue
    color10 = black on bright green
    color11 = black on bright magenta
    color12 = black on bright cyan
    color13 = black on bright yellow

verbose = on
```

And here is my `timewarrior.cfg`:
```
# Theme
import /opt/homebrew/Cellar/timewarrior/1.7.1/share/doc/timew/themes/dark.theme

define exclusions:
  monday = <8:30 12:00-13:00 >17:30
[...]

PR #738, opened 2026-01-30
Support coloring of tags with special characters

hype09:
As described in #591, adding coloring to a tag `tag-with-dashes` via config breaks the configuration parsing:

```
tags.tag-with-dashes.color = black on yellow
```

Additionally, anything that comes _after_ this line in the configuration file is no longer applied.

The reason for this lies in the way the configuration file is parsed, wrongly assuming that everything in front of the `=` is recognised as a single token by the lexer.

This pull request simplifies the parsing logic by checking if the current line contains a `=` symbol. If that's the case, the substrings before and after the symbol are treated as the assigned name and value. This also simplifies the case for settings with no value.

### DIFF
--- a/src/Rules.cpp
+++ b/src/Rules.cpp
@@ -372,22 +372,12 @@ void Rules::parse (const std::string& input, int nest /* = 1 */)
         }
         // Top-level settings:
         //   <name> '=' <value>
-        else if (tokens.size () >= 3 &&
-                 std::get <0> (tokens[1]) == "=")
+        //   <name> '='
+        else if (auto equals = line.find ('='); equals != std::string::npos)
         {
           // If this line is an assignment, then tokenizing it is a mistake, so use the raw data from 'line'.
-          auto equals = line.find ('=');
-          assert (equals != std::string::npos);
-
           set (trim (line.substr (indent, equals - indent)),
                trim (line.substr (equals + 1)));
-        }
-        // Top-level settings, with no value:
-        //   <name> '='
-        else if (tokens.size () == 2 &&
-                 std::get <0> (tokens[1]) == "=")
-        {
-          set (firstWord, "");
         }
         // Admit defeat.
         else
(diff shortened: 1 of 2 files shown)


lauft:
Closes #591

lauft (commented):
Looks good, however it seems to not work with the other config format.
When specifying colors like
```
define tags:
  SP+OC:
    color = white on red
  tag-with-dashes:
    color = blue on gray
```
the output of `timew sum` is uncolored... 🤔

hype09:
I did not know of that other format yet - I’ll have a look 👍

hype09:
Given the following configuration file content:

```
define tags:
  "Double quoted name":
    color = blue on gray
  'Single quoted name':
    color = red on white
  SP+OC:
    color = white on red
  SEPARATE WORDS:
    color = white on red
  tag-with-dashes:
    color = blue on gray
```

`timew config` will produce the following output:

```
define tags:
  "Double quoted name":
    color = blue on gray
  'Single quoted name':
    color = red on white
  OC:
    color = white on red
  WORDS:
    color = white on red
  dashes:
    color = blue on gray
```

I could do a similar change to what I did earlier, only for `Rules::parseGroup`, i.e. "take anything in front of the colon and trim the whitespace". This would help make the configuration parsing more predictable, but it's also a small BC break if people have things like `SEPARATE WORDS:` in their config and got accustomed to have it be parsed as `WORDS:`.

Do we need to be mindful of this or is it OK if the behavior for those edge cases changes to something more predictable?

lauft:
> Do we need to be mindful of this or is it OK if the behavior for those edge cases changes to something more predictable?

I guess the impact on existing workflows should be low (but [you never know](https://xkcd.com/1172)... 😅), so I opt for implementing this change 👍🏻

lauft:
Apart from other things keeping me busy, I also had to fix the issue with the openSUSE test images.
I rebased your branch to it and merged it.
Thanks again for your contribution and sorry for the delay.

Issue #483, opened 2022-03-09
Chart is displayed wrongly after a wide character


jyi2ya:
My data file is:

```plain
inc 20220309T093000Z - 20220309T110000Z # 测试
inc 20220309T110000Z - 20220309T120000Z # test
```

+ Actual behaviors

<img width="767" alt="shot0" src="https://user-images.githubusercontent.com/86813521/157365680-3bc00299-d654-4101-ad74-04425bfc108c.png">

+ Expected behavior

<img width="781" alt="shot1" src="https://user-images.githubusercontent.com/86813521/157365901-0315c92e-8bf2-4af1-b55b-a3aa76edf01d.png">

+ Diagnostics infomation

```plain

timew 1.4.3-dev
       Platform: Linux

Compiler:
        Version: 11.2.0
           Caps: +stdc +stdc_hosted +LP64 +c8 +i32 +l64 +vp64 +time_t64
     Compliance: C++11

Build Features
          Built: Mar  9 2022 08:56:40
         Commit: 8ba588dc
          CMake: 3.22.1
     Build type: 

Configuration
  TIMEWARRIORDB: -
            Cfg: /home/jyi/.timewarrior/timewarrior.cfg (-rw- 140 bytes)
       Database: /home/jyi/.timewarrior (drwx 4096 bytes)
        $EDITOR: vim
    Color theme: Built-in default

Extensions
       Location: /home/jyi/.timewarrior/extensions (drwx 4096 bytes)
               (None)
```

I have tried this code:

```cpp
#include "utf8.cpp"
#include <iostream>

int main()
{
	std::cout << utf8_text_width("测试") << std::endl;
	return 0;
}

```

The result was `4` as expected. So maybe there are some problem in timewarrior but not libshare.

TheTrueFlopsy:
[...]

PR #760, opened 2026-06-06
New test cases for chart display with non-ASCII tag names

TheTrueFlopsy:
I have added a few more test cases to `chart.t`. The intent is to provide evidence that the new version of `Composite` ([PR #117](/GothenburgBitFactory/libshared/pull/117) in libshared) fixes issue #483 and behaves as expected with some other kinds of non-ASCII text too. I have also included a couple of examples where the `timew {day,week,month}` charts are still not displayed as expected.


TheTrueFlopsy:
I have (almost) no idea why the test suite is failing on some platforms. On CentOS it's all the new `chart.t` test cases, but on the others it's the same three test cases in `summary.t`, even though I haven't touched that file. My spontaneous guesses are differences in behavior between Python versions, and/or my new `Composite` code breaking something in `timew summary`. But if it's the latter then it's strange that the problem is platform-dependent, and that no checks failed when the submodule reference was updated in `develop`. I'm using Debian trixie on my dev system and `summary.t` runs with zero failures there. I'll investigate what's going on and report my findings.

TheTrueFlopsy:
I think the issues that caused the failed checks have been dealt with now. The failure on CentOS was due to its version of Python not recognizing the `strict` keyword argument to `zip()`. The other failures happened because a single space was missing in some regular expressions in `summary.t`, and the resulting bug was only triggered when the test cases were run less than an hour after midnight, local system time.

test/chart.t
```diff
@@ -2,7 +2,8 @@
 
 ###############################################################################
 #
-# Copyright 2017 - 2019, Thomas Lauf, Paul Beckingham, Federico Hernandez.
+# Copyright 2017 - 2019, 2026, Thomas Lauf, Paul Beckingham,
+# Federico Hernandez, Johan Sarge.
```
lauft: Please change this to `Copyright 2017 - 2019, 2026, Gothenburg Bit Factory` (like in `summary.t`)
The explicit authorship mention in the header is phased out.

lauft:
@TheTrueFlopsy can you rebase your branch to the current `HEAD` on `develop`?
The issues with the test images should be fixed now.

TheTrueFlopsy:
Done and done. (I'm still working on the new line wrapper for libshared, by the way. Various unrelated things have attracted my attention, but progress is being made and it shouldn't take forever.)

lauft:
Unfortunately there are merge conflicts again. I am sorry, I should have merged this branch first... 😔

TheTrueFlopsy:
No problem. It should be OK again now.

lauft:
And merged. ✅

PR #765, opened 2026-07-20
remove dateutil dependency from totals.py

### LINKED ISSUE #764: ext/totals.py depends on third-party module
timewarrior 1.9.1 (and develop branch) with Python 3.13.13 on OpenSUSE Leap 16

----

The totals.py example report extension requires a third-party module, 'dateutil,' and I'm not sure why. The usage of it is simply replaced with calls to the standard library.

This is not ideal, as it means the report does not work out-of-the-box on a distribution-provided python installation.

I will submit a PR shortly to remedy this.

draeath:
Use datetime from the standard library, instead of the third-party module 'dateutil'

This allows totals.py to run on a standard python3 installation. Confirmed working with Python 3.13.

Fixes #764

### DIFF
--- a/ext/totals.py
+++ b/ext/totals.py
@@ -30,8 +30,6 @@ import datetime
 import json
 import sys
 
-from dateutil import tz
-
 DATEFORMAT = "%Y%m%dT%H%M%SZ"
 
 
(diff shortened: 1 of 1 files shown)


draeath:
will resubmit with DCO fix

PR #767, opened 2026-07-20
remove dateutil dependency from totals.py

draeath:
Use datetime from the standard library, instead of the third-party module 'dateutil'

This allows totals.py to run on a standard python3 installation. Confirmed working with Python 3.13.

Signed-off-by: Paul Bransford <draeath@gmail.com>

### DIFF
--- a/ext/totals.py
+++ b/ext/totals.py
@@ -30,8 +30,6 @@ import datetime
 import json
 import sys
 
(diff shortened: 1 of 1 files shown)


draeath:
this is infuriating, DCO doesn't understand I have multiple addresses

Issue #772, opened 2026-08-03
undo.data can be silently zero-filled by an unclean shutdown: AtomicFile never fsyncs before rename


tobixen:
I did run into issues with corrupt undo-log - twice over the last few weeks, partly due to an overgrown undo-log.  The analysis below is AI-generated.  I respect that some maintainers may want to close AI-generated issues and pull requests without spending time on explaining why.

---

⚠️ This comment is AI-generated (Claude Opus 5 via Claude Code) on behalf of tobixen

## Symptom

```
$ timew undo
Cannot handle line '<205 NUL bytes>txn:'
```

`undo.data` contains a run of NUL bytes glued directly onto a `txn:` marker, with no
newline between them. Once that is present, `timew undo` is unusable — it aborts on the
bad line rather than skipping it, so the entire undo feature is dead until the file is
repaired by hand.

This has happened to me twice in seven weeks — 2026-06-02 and 2026-07-23 — on
timewarrior 1.9.1, ext4 (`rw,relatime`), Linux 6.x/7.x.

## Why it goes unnoticed for weeks

On the default `journal.size = -1`, `timew undo` is the only command that *parses*
`undo.data`. Every other command appends to it by copying the existing bytes forward
verbatim. So the corruption is faithfully preserved and propagated by normal use, and you
only discover it the next time you happen to need `undo` — for me, 11 days after the fact.

(With `journal.size` set to a positive value this is very different, and worse — see
"Interaction with `journal.size`" below.)

## Root cause

`AtomicFile` is atomic with respect to *signals and interruption*, but not with respect
to *power loss*. In `src/AtomicFile.cpp`, `finalize_all()` closes the temp files and then
renames them into place with signals blocked:

```c++
  // Step 1: Close / Flush all the atomic files that may still be open. ...
  for (auto& file : impl::atomic_files)
  {
    file->close ();
  }
  ...
  // Step 2: Rename the temp files to the *real* file
  sigprocmask (SIG_SETMASK, &new_mask, &old_mask);
  for (auto& file : impl::atomic_files)
  {
    file->finalize ();
  }
  sigprocmask (SIG_SETMASK, &old_mask, nullptr);
```

There is no `fsync(2)` before the rename, and none after it on the containing directory.
In fact `fsync`, `fdatasync` and `O_SYNC` do not appear anywhere in `src/` (checked on
`develop` @ 8cda469a).

On ext4 with the default `data=ordered`, the rename can reach the journal while the data
[...]

PR #776, opened 2026-08-08
Update man page of 'undo' command

lauft:
Following suggestions in #772. Restructure the description, document the 'journal.size' configuration setting.

### DIFF
--- a/doc/man1/timew-undo.1.adoc
+++ b/doc/man1/timew-undo.1.adoc
@@ -9,11 +9,23 @@ timew-undo - revert Timewarrior commands
 
 == DESCRIPTION
 The 'undo' command is used to revert the action of Timewarrior commands.
-Only commands affecting intervals or Timewarrior configuration can be reverted.
+
 Timewarrior keeps a journal of changes to the interval database and Timewarrior configuration.
+Therefore, only commands affecting intervals or configuration can be reverted.
+
 A call to 'undo' removes the last entry in the journal and restores the previous state.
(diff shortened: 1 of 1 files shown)


doc/man1/timew-undo.1.adoc
```diff
@@ -9,12 +9,24 @@ timew-undo - revert Timewarrior commands
 
 == DESCRIPTION
 The 'undo' command is used to revert the action of Timewarrior commands.
-Only commands affecting intervals or Timewarrior configuration can be reverted.
+
 Timewarrior keeps a journal of changes to the interval database and Timewarrior configuration.
+Therefore, only commands affecting intervals or configuration can be reverted.
```
tobixen, line 14: The use of "only" here implies there exists commands doing other write-operations  affecting other things than the interval database and configuration.  Now I'm a bit curious what commands that may be - they should probably be listed explicitly or some example should be given.

lauft: The 'only' may be a bit overcautious from my side. 

Apart from intervals and configuration there is nothing else to manipulate in the Timewarrior DB. So if a command does not touch intervals or config, it touches nothing (internal).

The `report` command however may be an exception. A report can be any executable, and – although rather intended to be read-only – those could modify (external) files as well. So technically there is a file modification Timewarrior cannot undo.

For completeness, the commands that do not do journal entries are:
* chart
* diagnostics
* export
* extensions
* help
* get
* gaps
* report
* show
* summary
* tags

